PR #2, opened 2019-11-21
Actually working installation commands

baggepinnen:
Sorry for the noise. It turns out providing the url is only sufficient in the `pkg>` REPL mode. If using `using Pkg` one has to specify a `PackageSpec`


chriselrod:
Ah, yes, I should have caught that the first time. Thanks.

PR #12, opened 2020-01-08
Overhaul README with collapsible sections and add section discussing structs

MasonProtter:
The readme in its current form was a bit sprawling and hard to manage. I made sections of it collapsible so it is easier to navigate and get an overview. An alternative approach would be a table of contents. 

I also added a section called "Dealing with structs" where I recommend `StructArrays.jl` for kernels involving unsupported number types and demo it with that gemm kernel I posted on discourse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ function mul_avx!(C:: StructMatrixComplexFInt64, A::StructMatrixComplexFInt64, B
 end
 ```
 this `mul_avx!` kernel can now accept `StructArray` matrices of complex numbers and multiply them efficiently:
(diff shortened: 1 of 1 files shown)


chriselrod:
Thanks, this great!

I'll fix these typos on line 176, dealing with structs next time I make a commit:
"arrays conatining structs"
"currently reccomended that"

MasonProtter:
> I'll fix these typos on line 176, dealing with structs next time I make a commit:
"arrays conatining structs"
"currently reccomended that"

Whoops, sorry about that.

PR #24, opened 2020-01-20
broaden [compat]s, increment pkg version

JeffreySarnoff:


chriselrod:
Unless I accidentally fixed the problem, LoopVectorization will not actually work on Julia 1.0, so I left that bound as is.

It also definitely will not work on most tagged versions of VectorizationBase and SIMDPirates.
Within the next couple days, I'll set the versions of the dependencies and the bounds to "= 0.2" and fix the problem.

For now, I guess I'll let people face errors if the package manager will not actually put them on a new enough version.
For example, tests pass locally, but because I didn't update `Manifest.toml`, Travis CI fails.

PR #58, opened 2020-02-23
DO NOT MERGE: Delete Manifest.toml

DilumAluthge:


DilumAluthge:
@fredrikekre Any idea what is going on here? `SIMDPirates` 0.3.16 is in the General registry, so I’m not sure why Pkg thinks that 0.3.13 is the latest available version.

DilumAluthge:
See also https://github.com/JuliaLang/Pkg.jl/issues/1694

chriselrod:
Thank you for all the investigation and fixes.
Sounds like it has been resolved. Should I close this?

Issue #46, opened 2020-02-06
KeyError: key NumType not found


JeffreySarnoff:
This is a powerful and welcome tool.  I would like to apply it to more packages. I have run into a limitation on supported types that seems more narrow than necessary.

running e.g. `myselfdotavx(xs::T)` where T is a Float64 compatible primitive type (or most other numeric bitstype, including Int64 compatibles) except for T = Union{Float64, Float32, Int64, Int32} results in a `KeyError`.  
```
ERROR: KeyError: key XFloat32 not found
Stacktrace:
 [1] getindex at .\dict.jl:477 [inlined]
 [2] llvmtype(::Type{T} where T) at C:\Users\jas\.julia\packages\VectorizationBase\rPD9W\src\vectorizable.jl:22
 [3] #s13#2(::Any, ::Any, ::Any, ::Any) at C:\Users\jas\.julia\packages\VectorizationBase\rPD9W\src\VectorizationBase.jl:50
 [4] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any,N} where N) at .\boot.jl:526
 [5] vzero at C:\Users\jas\.julia\packages\VectorizationBase\rPD9W\src\VectorizationBase.jl:74 [inlined]
 [6] macro expansion at C:\Users\jas\.julia\packages\LoopVectorization\8wcli\src\reconstruct_loopset.jl:248 [inlined]
 [7] _avx_! at C:\Users\jas\.julia\packages\LoopVectorization\8wcli\src\reconstruct_loopset.jl:248 [inlined]
 [8] myselfdotavx(::Array{XFloat32,1}) at .\REPL[39]:3
 [9] ##core#495() at C:\Users\jas\.julia\packages\BenchmarkTools\7aqwe\src\execution.jl:297
 [10] ##sample#496(::BenchmarkTools.Parameters) at C:\Users\jas\.julia\packages\BenchmarkTools\7aqwe\src\execution.jl:303
 [11] _run(::BenchmarkTools.Benchmark{Symbol("##benchmark#494")}, ::BenchmarkTools.Parameters; verbose::Bool, pad::String, kwargs::Base.Iterators.Pairs{Symbol,Integer,NTuple{4,Symbol},NamedTuple{(:samples, :evals, :gctrial, :gcsample),Tuple{Int64,Int64,Bool,Bool}}}) at C:\Users\jas\.julia\packages\BenchmarkTools\7aqwe\src\execution.jl:331
[...]

Issue #19, opened 2020-01-13
Support for complex numbers?


hofmannmartin:
Does `LoopVectorization` support complex vectors? E.g.
```julia
using LoopVectorization

function test_avx!(x::Vector, y::Vector, beta)
  @avx for n=1:length(x)
    x[n] += conj(beta)*y[n]
  end
end

# set up
T = ComplexF32
N = 1024

x = zeros(T,N)
y = rand(T,N)
β = rand(T)
# end set up

test_avm!(x,y,β)
```
gives me the following error
```
KeyError: key Complex{Float32} not found

Stacktrace:
 [1] getindex at ./dict.jl:477 [inlined]
 [2] llvmtype(::Type) at /home/moeddel/.julia/packages/VectorizationBase/hBLD0/src/vectorizable.jl:21
 [3] #s24#123 at /home/moeddel/.julia/packages/SIMDPirates/aKvPC/src/memory.jl:94 [inlined]
 [4] #s24#123(::Any, ::Any, ::Any, ::Any, ::Any, ::Type, ::Any, ::Type, ::Any) at ./none:0
 [5] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any,N} where N) at ./boot.jl:524
 [6] vload at /home/moeddel/.julia/packages/SIMDPirates/aKvPC/src/memory.jl:218 [inlined]
 [7] macro expansion at ./gcutils.jl:91 [inlined]
 [8] test_avm!(::Array{Complex{Float32},1}, ::Array{Complex{Float32},1}, ::Complex{Float32}) at ./In[26]:4
```

chriselrod:
No, not yet.
Sometime after #17 we could add complex number support.

For now, I'd follow the [dealing-with-structs](https://github.com/chriselrod/LoopVectorization.jl#dealing-with-structs) approach from the README.

Note that `StructArray`s will be faster than the eventual support for arrays of structs.

lesshaste:
+1 for complex number support. It's also an opportunity to be faster than C compiled with gcc/clang which does a bad job of vectorizing complex arithmetic. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79102 , https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79336,  https://bugs.llvm.org/show_bug.cgi?id=31800 , https://bugs.llvm.org/show_bug.cgi?id=31677

hofmannmartin:
I managed to get it working
[...]

PR #65, opened 2020-03-04
Fix `for i in iter`

timholy:
This supports syntax like
```julia
for i in iter
    # do stuff
end
```
Because `iter` could be a `CartesianIndex` I presume we have to treat it the same way `eachindex(A)` gets treated.

Some of this is just whitespace changes (courtesy of Atom), but you can select to ignore it when reviewing. Also note I added a couple of `using` statements at the top of test files to facilitate running them independently. If you don't like that I can remove that part.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -31,6 +31,11 @@ uuid = "56ddb016-857b-54e1-b83d-db4d58db5568"
 deps = ["Base64"]
 uuid = "d6f4376e-aef5-505a-96c1-9c027394607a"
 
+[[OffsetArrays]]
+git-tree-sha1 = "707e34562700b81e8aa13548eb6b23b18112e49b"
+uuid = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
+version = "1.0.2"
+
 [[OrderedCollections]]
 deps = ["Random", "Serialization", "Test"]
 git-tree-sha1 = "c4c13474d23c60d20a67b217f1d7f22a40edf8f1"
@@ -49,9 +54,9 @@ uuid = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [[SIMDPirates]]
 deps = ["VectorizationBase"]
-git-tree-sha1 = "34dff4f4715f871e71b38f31397d96e62621f14d"
+git-tree-sha1 = "f91198b7ef74b04028f98e0eed7c556b93538a2e"
 uuid = "21efa798-c60a-11e8-04d3-e1a92915a26a"
-version = "0.6.5"
+version = "0.6.6"
 
 [[SLEEFPirates]]
 deps = ["Libdl", "SIMDPirates", "VectorizationBase"]
@@ -71,6 +76,6 @@ uuid = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [[VectorizationBase]]
 deps = ["CpuId", "LinearAlgebra"]
-git-tree-sha1 = "006d7b7f276db8d728f8bfd70ebf2efd132f9548"
+git-tree-sha1 = "8abb5697fb64cadccd1bba444c955942d3181e5c"
 uuid = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
-version = "0.7.0"
+version = "0.7.1"
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.6.20"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
 SIMDPirates = "21efa798-c60a-11e8-04d3-e1a92915a26a"
 SLEEFPirates = "476501e8-09a2-5ece-8869-fb82de89a1fa"
--- a/src/LoopVectorization.jl
+++ b/src/LoopVectorization.jl
@@ -12,6 +12,7 @@ using SIMDPirates: VECTOR_SYMBOLS, evadd, evmul, vrange, reduced_add, reduced_pr
     vmullog2, vmullog10, vdivlog2, vdivlog10, vmullog2add!, vmullog10add!, vdivlog2add!, vdivlog10add!, vfmaddaddone
 using Base.Broadcast: Broadcasted, DefaultArrayStyle
 using LinearAlgebra: Adjoint, Transpose
+using Base.Meta: isexpr
 
 const SUPPORTED_TYPES = Union{Float16,Float32,Float64,Integer}
 
--- a/src/LoopVectorization.jl
+++ b/src/LoopVectorization.jl
@@ -4,7 +4,7 @@ using VectorizationBase, SIMDPirates, SLEEFPirates, Parameters
 using VectorizationBase: REGISTER_SIZE, REGISTER_COUNT, extract_data, num_vector_load_expr,
     mask, masktable, pick_vector_width_val, valmul, valrem, valmuladd, valadd, valsub, _MM,
     maybestaticlength, maybestaticsize, staticm1, subsetview, vzero, stridedpointer_for_broadcast,
-    Static, StaticUnitRange, StaticLowerUnitRange, StaticUpperUnitRange,
+    Static, StaticUnitRange, StaticLowerUnitRange, StaticUpperUnitRange, unwrap, maybestaticrange,
     PackedStridedPointer, SparseStridedPointer, RowMajorStridedPointer, StaticStridedPointer, StaticStridedStruct
 using SIMDPirates: VECTOR_SYMBOLS, evadd, evmul, vrange, reduced_add, reduced_prod, reduce_to_add, reduce_to_prod,
     sizeequivalentfloat, sizeequivalentint, vadd!, vsub!, vmul!, vfdiv!, vfmadd!, vfnmadd!, vfmsub!, vfnmsub!,
@@ -22,6 +22,7 @@ export LowDimArray, stridedpointer, vectorizable,
     vfilter, vfilter!
 
 
+include("vectorizationbase_extensions.jl")
 include("map.jl")
 include("filter.jl")
 include("costs.jl")
--- a/src/add_loads.jl
+++ b/src/add_loads.jl
@@ -70,13 +70,3 @@ function add_loopvalue!(ls::LoopSet, arg::Symbol, elementbytes::Int)
     loopsymop
 end
 
-
-struct LoopValue end
-@inline VectorizationBase.stridedpointer(::LoopValue) = LoopValue()
-@inline VectorizationBase.vload(::LoopValue, i::Tuple{_MM{W}}) where {W} = _MM{W}(@inbounds(i[1].i) + 1)
-# @inline VectorizationBase.vload(::LoopValue, i::Tuple{_MM{W}}, ::Unsigned) where {W} = _MM{W}(@inbounds(i[1].i) + 1)
-@inline VectorizationBase.vload(::LoopValue, i::Tuple{_MM{W}}, ::Mask) where {W} = _MM{W}(@inbounds(i[1].i) + 1)
-@inline VectorizationBase.vload(::LoopValue, i::Integer) = i + one(i)
-@inline VectorizationBase.vload(::LoopValue, i::Tuple{I}) where {I<:Integer} = @inbounds(i[1]) + one(I)
-@inline Base.eltype(::LoopValue) = Int8
-
--- a/src/graphs.jl
+++ b/src/graphs.jl
@@ -33,7 +33,7 @@
 
 # For passing options like array types and mask
 # struct LoopSetOptions
-    
+
 # end
 
 struct Loop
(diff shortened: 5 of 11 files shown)


timholy:
One cautionary note: in a test script [here](https://gist.github.com/timholy/7b0d9bc2fc95af01fba182e7f0095a8b) I get this:
```julia
julia> @btime old2d!($out, $A, $kern);
  67.972 μs (0 allocations: 0 bytes)

julia> @btime avx2d!($out, $A, $kern);
  787.508 μs (0 allocations: 0 bytes)
```
It seems possible in principle that this is a consequence of this PR.

As you'll see in the gist, other challenges include:
- the use of OffsetArrays
- a lot of code is generic multidimensional
- element types include `N0f8` from the FixedPointNumbers package, which is a mapping of `0x00:0xff`->`0.0:1/255:1.0` (i.e., 8-bit values that span the range 0.0 to 1.0, inclusive).

timholy:
Update: yep, that's a lot of it but not all of it: I added an `avx2d_unpack!` which converts those into UnitRanges and I get
```julia
julia> @btime avx2d_unpack!($out, $A, $kern);
  101.279 μs (0 allocations: 0 bytes)
```
which is much better but still worse than the `@avx`-free version. CC @stillyslalom.

chriselrod:
```julia
# include your gist here
using VectorizationBase, BenchmarkTools
# gesp is get element strided pointer
VectorizationBase.stridedpointer(A::OffsetArray) = VectorizationBase.gesp(stridedpointer(parent(A)), -1 .* A.offsets)

function avx2d_unpackall!(out::AbstractMatrix, A::AbstractMatrix, kern::OffsetArray, R=CartesianIndices(out), z=zero(eltype(out)))
    rng1k, rng2k = axes(kern)
    rng1kf, rng1kl = first(rng1k), last(rng1k)
    rng2kf, rng2kl = first(rng2k), last(rng2k)
    rng1,  rng2  = R.indices
    rng1f, rng1l = first(rng1), last(rng1)
    rng2f, rng2l = first(rng2), last(rng2)
    # Manually unpack the OffsetArray
    kernA = parent(kern)
    o1, o2 = kern.offsets
    @avx for j in rng2f:rng2l, i in rng1f:rng1l
        tmp = z
        for jk in rng2kf:rng2kl, ik in rng1kf:rng1kl
            tmp += A[i+ik,j+jk]*kernA[ik-o1,jk-o2]
        end
        out[i,j] = tmp
    end
    out
end

A = rand(Float32, 100, 100);
kern = centered(rand(Float32, 3, 3));
out1 = OffsetArray(similar(A, size(A).-2), 1, 1);   # stay away from the edges of A
out2 = similar(out1); out3 = similar(out1);

@benchmark old2d!($out1, $A, $kern)
@benchmark avx2d_unpack!($out2, $A, $kern)
@benchmark avx2d_unpackall!($out3, $A, $kern)

out1 ≈ out2
out1 ≈ out3
```
On my computer (with avx512), this yields
```julia
julia> out1 = OffsetArray(similar(A, size(A).-2), 1, 1);   # stay away from the edges of A

julia> out2 = similar(out1); out3 = similar(out1);

julia> @benchmark old2d!($out1, $A, $kern)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     66.879 μs (0.00% GC)
  median time:      66.952 μs (0.00% GC)
  mean time:        67.110 μs (0.00% GC)
  maximum time:     107.237 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark avx2d_unpack!($out2, $A, $kern)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     109.412 μs (0.00% GC)
  median time:      109.645 μs (0.00% GC)
  mean time:        109.798 μs (0.00% GC)
  maximum time:     156.803 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark avx2d_unpackall!($out3, $A, $kern)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     3.055 μs (0.00% GC)
  median time:      3.074 μs (0.00% GC)
  mean time:        3.078 μs (0.00% GC)
  maximum time:     5.444 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     8

julia> out1 ≈ out2
true

julia> out1 ≈ out3
true
```

Applying `@avx` to the outermost loop resulted in a roughly 20x performance gain.

Normally, applying `@avx` to the outermost loop should yield the greatest benefit, as this gives it more opportunity to get creative with nested loop optimizations.

It may be worth considering passing size information when some loops are much longer than others (eg, the outer loops vs the inner loops) to reduce the risk of it deciding to do something unproductive. E.g., it could decide to 4x unroll and vectorize (8x or 16x factor for single precision depending on avx/avx512), yielding a total of 32x or 64x unrolling. If one loop is only 7 iterations, it had best not do that.

I added "size hint" infrastructure that there currently isn't any way to take advantage of. These are used for optimization decisions, but not for lowering. Currently, the hint is always set to 1024. Perhaps we can use this in some way to make sure it optimizes as though these inner loops are short, while still keeping the number of iterations dynamic.

> challenges include: the use of OffsetArrays

I should add that offset array specialization for `stridedpointer` somewhere. Thankfully, extending support to new array types is fairly straightforward so long as you can get a pointer to the memory and they are strided.

> challenges include: a lot of code is generic multidimensional

It would be fantastic to add `CartesianIndices` support. 
The `@avx` macro calls a generated function. This is used to generate different code depending on whether or not arrays are transposed, as loop order is likely to change as a result. (`@_avx` is a more naive macro that immediately emits code, making generic assumptions.)
Similarly, it would be possible to make the number of loops change based on the type of the `CartesianIndices`.
The trick would be substituting the associated index and replacing it with multiple on each `ArrayReference`.
It would be great to have this, so I'll probably start working on it fairly soon.

> challenges include: element types include `N0f8` from the FixedPointNumbers package, which is a mapping of 

Two problems here. The first is that support is currently lousy for mixing types within the loops. I originally envisioned everything being `Float64`. I quickly realized that it would be so easy to support other types that it would be silly not to, but extensions from there have all been hacks that haven't really addressed the problem of considering things can be different types.
This should all be redone eventually; while I've fixed almost all the issues that have come up, hacks are brittle and much more likely to crack than an engineered approach.

The second problem -- supporting generic number types -- is much harder. 
For anyone reading and wanting a summary of the problem: LoopVectorization uses `NTuple{N,Core.VecElement{T}}`, where `T` is something that lets this map to an LLVM vector rather than an LLVM array. LLVM vectors have lots of operations defined on them, many of which map closely to assembly instructions on many architectures, and are how it tells LLVM exactly what it needs to do.

The problem then is that given type `foo`, and operations a user has defined on `foo` like `+` and `/`, I need to be able to somehow construct these `NTuple{N,Core.VecElement{T}}`s, and also need to know what the `+` and `/` are supposed to do in terms of primitive types. 

[Here](https://github.com/chriselrod/LoopVectorization.jl/issues/19) is an issue specific to `Complex` numbers, and [here](https://github.com/chriselrod/LoopVectorization.jl/issues/46) is an issue about supporting generic types. I also discussed it [briefly at the end of this comment](https://discourse.julialang.org/t/ann-loopvectorization/32843/86?u=elrod), but I don't have any good ideas.

The manual approach of adding support per element type would work, but I am sure there is a means aligned with and showing off Julia's generic power.
Perhaps I could define `SVec`, the wrapper type for `NTuple{W,Core.VecElement{T}}` as `<:Number`. Then a lot of operations would work, as they could sit inside the fields of the struct if they allow holding generic numbers. But then some work is needed to figure out how to load and store elements from arrays of structs.

I'll take a look at what's going wrong with the iterables performance-wise soon.

timholy:
Love it!

> >    challenges include: element types include N0f8 from the FixedPointNumbers package, which is a mapping of

> Two problems here. The first is that support is currently lousy for mixing types within the loops...
> The second problem -- supporting generic number types -- is much harder...I need to be able to somehow construct these NTuple{N,Core.VecElement{T}}s, and also need to know what the + and / are supposed to do in terms of primitive types.

The ultimate solution seems to be moving this into the compiler, since there you have everything (you know all the types). But in the meanwhile, would it help to use https://github.com/JuliaArrays/TiledIteration.jl and convert chunks of the array to a common numeric type?

timholy:
As an example of your unrolling, it would be nice for handling colors:
```julia
using ImageCore
c = RGB(0.12, 0.45, 0.38)
```
Typically the filter is real-valued but pixels are often color.

chriselrod:
I'm going to push to this PR soon. I'm changing how loop bounds are handled.

chriselrod:
> The ultimate solution seems to be moving this into the compiler, since there you have everything (you know all the types). 

Ultimately, it would be great if MLIR (or another project) renders the performance side of LoopVectorization obsolete. But punting the loop to a generated function does give us type information useful for some things, but not nearly as much information as available to the compiler.

> But in the meanwhile, would it help to use https://github.com/JuliaArrays/TiledIteration.jl and convert chunks of the array to a common numeric type?

What sort of type conversions can it do?
Many examples should work fine as is.
However, for some problems, TiledIteration would be worth using for the sake of tiling the loops into cache-friendly array sizes, and for it's threading support. By (recursively) dividing matrices into 64x64 blocks and then applying a LoopVectorization-based kernel, Gaius.jl was about 3x faster in single threaded mode on my computer than applying the kernels directly to 1000x1000 matrices.
I think the advantage will be smaller for image filtering problems, because you're not forced to pass over array `A` O(N) times, but depending on the kernels, 

In my recent commits, I now have `eachindex` (and the symbol fallback) always check the `first` and `last`.
Additionally, I realized that while `eachindex` on `OffsetArray`s of dimension 1 returns its offsets, when the `OffsetArray` is of dimension 2 or greater, `eachindex` returns `OneTo`. Therefore, whenever the dimension is 2 or greater, offsetting the stridedpointer as I did above would only work if someone isn't `eachindex`ing the array.

timholy:
No rush, I'm playing with an an attempt to add support for external passes to Julia's compiler. The idea is to allow you to set up a callback function that receives the type-inferred CodeInfo with something like
```julia
macro avx(ex)
    esc(Expr(:block, Expr(:meta, Expr(:external_pass, :start, avx_pass)), ex, Expr(:meta, Expr(:external_pass, :stop))))
end
```
where you've defined the function `avx_pass` to take a [`Core.Compiler.OptimizationState`](https://github.com/JuliaLang/julia/blob/4d9a3349170d9beea8584cc9a730d37609280ac3/base/compiler/optimize.jl#L7-L18).

If this works you'll really hate me because it would basically encourage you to rewrite this package from scratch :frowning:. OTOH in the long run it should make a lot of things a lot easier.

chriselrod:
I don't seem to know how to push to PRs; I pushed here by mistake:
https://github.com/chriselrod/LoopVectorization.jl/tree/teh/loops
Now:
```julia
julia> @benchmark old2d!($out2, $A, $kern)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     67.172 μs (0.00% GC)
  median time:      67.442 μs (0.00% GC)
  mean time:        67.769 μs (0.00% GC)
  maximum time:     111.490 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark avx2d!($out2, $A, $kern)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     130.930 μs (0.00% GC)
  median time:      132.447 μs (0.00% GC)
  mean time:        132.576 μs (0.00% GC)
  maximum time:     188.808 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark avx2douter!($out3, $A, $kern)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.662 μs (0.00% GC)
  median time:      2.684 μs (0.00% GC)
  mean time:        2.686 μs (0.00% GC)
  maximum time:     4.959 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     9
```
Where they're [defined here](https://github.com/chriselrod/LoopVectorization.jl/blob/teh/loops/test/offsetarrays.jl) (and tests pass), using ranges for iteration.

timholy:
> Therefore, whenever the dimension is 2 or greater, offsetting the stridedpointer as I did above would only work if someone isn't eachindexing the array.

If it's a `IndexLinear()` array (which holds for any `OffsetArray` wrapper around any other kind of `IndexLinear()` array), and they want cartesian indices, they should be doing `for I in CartesianIndices(A)`.

timholy:
> I don't seem to know how to push to PRs; I pushed here by mistake: https://github.com/chriselrod/LoopVectorization.jl/tree/teh/loops

I don't pretend to really understand this package, but FWIW it looks good to me! If you're happy merge whenever. Thanks for the amazing support! I'll keep you posted about the compiler work.

chriselrod:
I can merge this and that commit separately, or I can wait until you merge that commit into this PR 
https://github.com/timholy/LoopVectorization.jl/pull/1
and then merge it.

On indexing `OffsetArray`, the problem was that it uses offsets when indexed with cartesian or multiple indexes, but not indexed linearly. LoopVectorization uses fat pointers (that carry their strides) internally for indexing. My original idea was to offset the base of the pointer to compensate for the OffsetArray's offset and return an ordinary stridedpointer, but this wouldn't show the different behaviors in different situations.
So I figured the simplest approach was to just introduce a new type with the same behavior pattern.

Thanks for the PR and interest. I'll definitely be interested in your compiler work, especially if like `@code_typed` it can strip away most of Julia's abstractions, thus solving the arbitrary `struct` problem. I'll just have to make sure I don't lose my job for playing with loops all day.

timholy:
Somehow it's not updating that PR. Feel free to just open a PR at that `teh/loops` branch if that's easier.

PR #69, opened 2020-03-07
Add OffsetArray tests for 1d

timholy:
This is basically janitorial work as I am starting to systematically read through the entire `src/` directory to get a better understanding of how this works. The bulk of the line changes here just come from adding an additional layer of `@testset` depth (none of the code in the `2d` case changed), the only contribution here other than code-comments is the new `1d` tests.

### DIFF
--- a/src/vectorizationbase_extensions.jl
+++ b/src/vectorizationbase_extensions.jl
@@ -17,8 +17,10 @@ struct OffsetStridedPointer{T, N, P <: VectorizationBase.AbstractStridedPointer{
     ptr::P
     offsets::NTuple{N,Int}
 end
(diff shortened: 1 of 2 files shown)


chriselrod:
I had added one-dimensional offset array tests to the [dot tests](https://github.com/chriselrod/LoopVectorization.jl/blob/master/test/dot.jl#L193), which would be fine to remove now.

The tests could eventually use some rework to make them more orthogonal. But with several axis, it isn't clear how to organize them (e.g. testing reductions, vs testing Offset Arrays). `dot` and `gemm` probably aren't that useful from an organizational standpoint, other than them being specific things I'd like to keep testing.

They take a while to run (10-15 minutes on Travis).

timholy:
Ah, sorry I missed those. No, let's close this instead.

PR #70, opened 2020-03-08
Add a few more docstrings

timholy:
Other than `_avx_!`, every export is documented including `LoopVectorization` itself. (I don't always remember, but I often like to provide a brief API summary of a package in the module's docstring). It also adds a new API reference page to the Documenter docs. 

Sorry about the whitespace changes, my Atom settings do it automatically.


chriselrod:
These look great.

PR #71, opened 2020-03-08
Add a few convenience utilities

timholy:
The main advantage these new predicates have is a small improvement in performance due to not needing to construct new objects for comparisons.

For Dicts of types, this switches to IdDict. This is both faster (particularly on construction), and safe for precompilation (see  https://docs.julialang.org/en/latest/manual/modules/#Module-initialization-and-precompilation-1).


chriselrod:
This looks great, thanks for these improvements!
I wasn't aware of the possible danger of using a `Dict` for `DataType`s or `Function`s.

PR #75, opened 2020-03-16
Preserve names of arrays and iteration variables across generated boundary

### LINKED ISSUE #74: A simpler DSL? (Pass code literal to _avx_!)
It occurred to me that one thing making it a bit hard to debug problems is that in the final generated code, the names of the variables are mostly unrelated to the original code. Have you considered passing the raw expression as a type parameter to `_avx_!`? I'm envisioning something like translating
```julia
ex = :(for i = 1:n
    s += a[i]
end)
```
into
```julia
struct TypedExpr{head,args} end
struct TailTuple{head,tail} end
TypedExpr{:for,TailTuple{TypedExpr{:(=),TailTuple{:i,TypedExpr{:call,TailTuple{:colon,TailTuple{1,:n}}}...
```
This is basically just an `Expr` but one that could be passed as a type-parameter.

If you like this idea I can work on it.

timholy:
Fixes #74

### DIFF
--- a/src/condense_loopset.jl
+++ b/src/condense_loopset.jl
@@ -4,11 +4,13 @@
 Base.:|(u::Unsigned, it::IndexType) = u | UInt8(it)
 Base.:(==)(u::Unsigned, it::IndexType) = (u % UInt8) == UInt8(it)
 
-struct ArrayRefStruct
+struct ArrayRefStruct{array,ptr}
     index_types::UInt64
     indices::UInt64
     offsets::UInt64
 end
+array(ar::ArrayRefStruct{a,p}) where {a,p} = a
+ptr(ar::ArrayRefStruct{a,p}) where {a,p}   = p
 
 function findindoradd!(v::Vector{T}, s::T) where {T}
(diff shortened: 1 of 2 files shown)


chriselrod:
This is fantastic!

As a quick comment, I avoided making any of the structs parametric because I wanted to move towards everything being concretely typed as quickly as possible, although I haven't really dived into how well I'm actually achieving that. Regardless, `gensym` is actually much slower than a dynamic dispatch, so this should speed up compilation if anything...

timholy:
If we add enough `@nospecialize` statements everywhere, concrete typing probably shouldn't matter. (You don't have any reason other than compile performance to want that, right?)

chriselrod:
Yes, compile performance is the only reason. Any general advice on using `@nospecialize`?

timholy:
I'll submit a PR. Best approach is to analyze the number of specializations per method; that tells you how many times they've been compiled. Whole goal is to avoid recompilation.

timholy:
Not as straightforward as I hoped, it may be a Julia bug. https://github.com/JuliaLang/julia/issues/35131

PR #76, opened 2020-03-16
WIP: shorten compile time with nospecialize

timholy:
This is a branch that, for the moment, is mostly for https://github.com/JuliaLang/julia/issues/35131. Steps to reproduce: after `dev`ing this branch, do

```julia
julia> using LoopVectorization

shell> pwd
/home/tim/.julia/dev/LoopVectorization/test

julia> include("broadcast.jl")
(T, #= /home/tim/.julia/dev/LoopVectorization/test/broadcast.jl:7 =# @__LINE__()) = (Float32, 7)
(T, #= /home/tim/.julia/dev/LoopVectorization/test/broadcast.jl:7 =# @__LINE__()) = (Float64, 7)
(T, #= /home/tim/.julia/dev/LoopVectorization/test/broadcast.jl:7 =# @__LINE__()) = (Int32, 7)
(T, #= /home/tim/.julia/dev/LoopVectorization/test/broadcast.jl:7 =# @__LINE__()) = (Int64, 7)
Test Summary: | Pass  Total
broadcast     |   74     74
Test.DefaultTestSet("broadcast", Any[], 74, false)

julia> include("/home/tim/src/julia/nspecializations.jl")   # just if you want a quick count
allmethods (generic function with 1 method)

julia> nspecializations(first(methods(LoopVectorization.abstractparameters)))
33
```
To be clear, the point of `abstractparameters` is to allow one to write certain methods without a `where` clause and still be able to access the parameters. I am assuming that having the `where` will force specialization independent of a `@nospecialize`. (If that's not correct, I'd rather get rid of `abstractparameters` and go back to the way they were before.) So it's ironic that `abstractparameters` itself seems to get inferred for many different input types.

### DIFF
--- a/src/LoopVectorization.jl
+++ b/src/LoopVectorization.jl
@@ -25,6 +25,7 @@ export LowDimArray, stridedpointer, vectorizable,
 
 include("vectorizationbase_extensions.jl")
 include("predicates.jl")
+include("typeutils.jl")
 include("map.jl")
 include("filter.jl")
 include("costs.jl")
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -60,8 +60,9 @@ the final operation.
 # TODO: Need to make this handle A or B being (1 or 2)-D broadcast objects.
 function add_broadcast!(
     ls::LoopSet, mC::Symbol, bcname::Symbol, loopsyms::Vector{Symbol},
-    ::Type{Product{A,B}}, elementbytes::Int
-) where {A, B}
+    @nospecialize(AB::Type{<:Product}), elementbytes::Int
+)
+    A, B = AB.parameters
     K = gensym(:K)
     mA = gensym(:Aₘₖ)
     mB = gensym(:Bₖₙ)
@@ -126,8 +127,9 @@ function LowDimArray{D}(data::A) where {D,T,N,A <: AbstractArray{T,N}}
 end
 function add_broadcast!(
     ls::LoopSet, destname::Symbol, bcname::Symbol, loopsyms::Vector{Symbol},
-    ::Type{<:LowDimArray{D,T,N}}, elementbytes::Int
-) where {D,T,N}
+    @nospecialize(LDA::Type{<:LowDimArray}), elementbytes::Int
+)
+    D, T, N = LDA.parameters
     fulldims = Symbol[loopsyms[n] for n ∈ 1:N if D[n]]
     ref = ArrayReference(bcname, fulldims)
     add_simple_load!(ls, destname, ref, elementbytes, true, false )::Operation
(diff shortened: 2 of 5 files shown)


chriselrod:
Running the test suite prints a lot of times. How noisy are Travis runs?

I can also run the tests locally to compare as well.

timholy:
Do you mean, can one use the time as a guide to progress? Or, might I have left a stray `@show` or something in a previous PR? I get a lot of output too.

timholy:
Actually, this might already be helping: if memory serves it used to be ~205s for me to run the whole test suite, this run came out as 177. Of course that's completely uncontrolled (what else was my machine doing at the same time? was I plugged into power?), but something to keep an eye on.

chriselrod:
Yes, I was wondering if we could use Travis to track progress.
The [latest run on master](https://travis-ci.com/github/chriselrod/LoopVectorization.jl/jobs/298554607), Julia 1.1, Linux: 687 seconds
[This PR](https://travis-ci.com/github/chriselrod/LoopVectorization.jl/jobs/298716084), Julia 1.1, Linux: 666 seconds

timholy:
Here's results from a more controlled experiment where I ran each twice:
- master: 
  + 213.179568
  + 211.746881
- this PR: 
  + 212.729866
  + 214.747268

I don't know what happened with that one fast run, but so far this isn't helping. Of course, if we've identified a crack in the control of specializations (which I suspect we have), the conclusion might change.

These tests were done on Julia 1.3.

chriselrod:
Would it be better to ditch [the precompile file](https://github.com/chriselrod/LoopVectorization.jl/blob/master/src/precompile.jl) when timing this? That file shows a huge list of methods from the broadcast file I don't want it specializing on, such as specific static sizes or patterns of true/false in LowDimArray.

On a couple day old build of Julia master, I got, with LoopVectorization master
255.619275 seconds (743.55 M allocations: 32.386 GiB, 2.51% gc time)
255.793652 seconds (743.55 M allocations: 32.386 GiB, 2.61% gc time)
255.626683 seconds (743.55 M allocations: 32.386 GiB, 2.59% gc time)

This PR, after merging master
253.579559 seconds (739.90 M allocations: 32.215 GiB, 2.64% gc time)
253.793647 seconds (739.90 M allocations: 32.215 GiB, 2.61% gc time)
252.885642 seconds (739.90 M allocations: 32.215 GiB, 2.59% gc time)

I've disabled all turboing on this desktop (i.e., it'll run at the same speed regardless of how many cores are under load, or how long it has been working [although speeds will change as a function of work, so clocks vary with sse/avx/avx512/idle, but that should be consistent within a given workload]; but that does mean that the max speed it I have it configured to run at is actually 0.2 GHz slower than the stock max single core boost).

There's still a fair bit of noise, but this PR does already look a bit faster on my computer. It seems to allocate a little less memory.

timholy:
Great observation. This PR is very much partial, so it might get even better. Still, I have a sense that something isn't right, so I'll wait a bit and see if Jeff discovers anything else.

timholy:
Superseded by #97

PR #84, opened 2020-04-05
WIP: support mixed Int/CartesianIndex indexing

timholy:
ImageFiltering contains optimizations for separable filters: if the filter size along each dimension is `n[i]`, then per-pixel a fully-separable filter has cost `O(n[1] + ... + n[d])`, where as a non-separable filter has cost `O(n[1] * ... * n[d])`. Consequently this can be a very large savings. This PR aims to support such operations by extending support for the exponential-filter demo in https://julialang.org/blog/2016/02/iteration/.

This doesn't yet work, because I haven't yet compensated for what the commented-out `@assert` is supposed to catch. I thought I'd submit this now to discuss the right architecture changes. My presumption is that the easiest approach is to return `Δidxs` from `calcnops` and thus make `nopsv` a `Vector{Vector{Int}}`. But if you have a better suggestion I am all ears.

I'll leave a couple of comments in the changes.

### DIFF
--- a/src/condense_loopset.jl
+++ b/src/condense_loopset.jl
@@ -113,23 +113,23 @@ function OperationStruct!(varnames::Vector{Symbol}, ls::LoopSet, op::Operation)
 end
 ## turn a LoopSet into a type object which can be used to reconstruct the LoopSet.
 
+function loop_boundary(loop::Loop)
+    startexact = loop.startexact
+    stopexact = loop.stopexact
+    if startexact & stopexact
+        Expr(:call, Expr(:curly, lv(:StaticUnitRange), loop.starthint, loop.stophint))
+    elseif startexact
+        Expr(:call, Expr(:curly, lv(:StaticLowerUnitRange), loop.starthint), loop.stopsym)
+    elseif stopexact
+        Expr(:call, Expr(:curly, lv(:StaticUpperUnitRange), loop.stophint), loop.startsym)
+    else
+        Expr(:call, :(:), loop.startsym, loop.stopsym)
+    end
+end
 
 function loop_boundaries(ls::LoopSet)
     lbd = Expr(:tuple)
-    for loop ∈ ls.loops
-        startexact = loop.startexact
-        stopexact = loop.stopexact
-        lexpr = if startexact & stopexact
-            Expr(:call, Expr(:curly, lv(:StaticUnitRange), loop.starthint, loop.stophint))
-        elseif startexact
-            Expr(:call, Expr(:curly, lv(:StaticLowerUnitRange), loop.starthint), loop.stopsym)
-        elseif stopexact
-            Expr(:call, Expr(:curly, lv(:StaticUpperUnitRange), loop.stophint), loop.startsym)
-        else
-            Expr(:call, :(:), loop.startsym, loop.stopsym)
-        end
-        push!(lbd.args, lexpr)
-    end
+    foreach(loop -> push!(lbd.args, loop_boundary(loop)), ls.loops)
     lbd
 end
 
@@ -204,6 +204,7 @@ end
     ::Val{UT}, ::Type{OPS}, ::Type{ARF}, ::Type{AM}, ::Type{LPSYM}, lb::LB,
     ::Val{AR}, ::Val{D}, ::Val{IND}, subsetvals, arraydescript, vargs::Vararg{<:Any,N}
 ) where {UT, OPS, ARF, AM, LPSYM, LB, N, AR, D, IND}
+    1 + 1
     num_vptrs = length(ARF.parameters)::Int
     vptrs = [gensym(:vptr) for _ ∈ 1:num_vptrs]
     call = Expr(:call, lv(:_avx_!), Val{UT}(), OPS, ARF, AM, LPSYM, :lb)
@@ -279,14 +280,23 @@ function generate_call(ls::LoopSet, IUT, debug::Bool = false)
     add_external_functions!(q, ls)
     q
 end
-
+concat_vals() = Val{()}()
+# @generated concat_vals(::Val{N}) where {N} = Val{(N,)}()
+# @generated concat_vals(::Val{M}, ::Val{N}) where {M, N} = Val{(M,N)}()
+@generated function concat_vals(args...)
+    tup = Expr(:tuple)
+    for n in eachindex(args)
+        push!(tup.args, args[n].parameters[1])
+    end
+    Expr(:call, Expr(:curly, :Val, tup))
+end
 function setup_call_noinline(ls::LoopSet, U = zero(Int8), T = zero(Int8))
     call = generate_call(ls, (false,U,T))
     hasouterreductions = length(ls.outer_reductions) > 0
     q = Expr(:block)
     vptrarrays = Expr(:tuple)
     vptrsubsetvals = Expr(:tuple)
-    vptrsubsetdims = Expr(:tuple)
+    vptrsubsetdims = Expr(:call, lv(:concat_vals))
     vptrindices = Expr(:tuple)
     stridedpointerLHS = Symbol[]
     loopvalueLHS = Symbol[]
@@ -304,7 +314,7 @@ function setup_call_noinline(ls::LoopSet, U = zero(Int8), T = zero(Int8))
                         @assert array ∈ loopvalueLHS
                         push!(vptrarrays.args, -1)
                     end
-                    push!(vptrsubsetdims.args, nothing)
+                    push!(vptrsubsetdims.args, Expr(:call, Expr(:curly, :Val, nothing)))
                     vp = first(ex.args)::Symbol
                     push!(stridedpointerLHS, vp)
                     push!(vptrindices.args, findfirst(a -> vptr(a) == vp, ls.refs_aliasing_syms))
@@ -316,7 +326,7 @@ function setup_call_noinline(ls::LoopSet, U = zero(Int8), T = zero(Int8))
                         @assert vptrarrayid isa Int
                     end
                     push!(vptrarrays.args, vptrarrayid::Int)
-                    push!(vptrsubsetdims.args, ex.args[2].args[3].args[1].args[2])
+                    push!(vptrsubsetdims.args, ex.args[2].args[3])#.args[1].args[2])
                     push!(vptrsubsetvals.args, ex.args[2].args[4])
                     vp = first(ex.args)::Symbol
                     push!(stridedpointerLHS, vp)
@@ -327,7 +337,8 @@ function setup_call_noinline(ls::LoopSet, U = zero(Int8), T = zero(Int8))
         push!(q.args, ex)
     end
     insert!(call.args, 8, Expr(:call, Expr(:curly, :Val, vptrarrays)))
-    insert!(call.args, 9, Expr(:call, Expr(:curly, :Val, vptrsubsetdims)))
+    # insert!(call.args, 9, Expr(:call, Expr(:curly, :Val, vptrsubsetdims)))
+    insert!(call.args, 9, vptrsubsetdims)
     insert!(call.args, 10, Expr(:call, Expr(:curly, :Val, vptrindices)))
     insert!(call.args, 11, vptrsubsetvals)
     if hasouterreductions
--- a/src/memory_ops_common.jl
+++ b/src/memory_ops_common.jl
@@ -29,17 +29,46 @@ function add_vptr!(ls::LoopSet, array::Symbol, vptrarray::Symbol = vptr(array), 
     end
     nothing
 end
-function subset_vptr!(ls::LoopSet, vptr::Symbol, indnum::Int, ind::Union{Symbol,Int})
+
+@inline valsum() = Val{0}()
+@inline valsum(::Val{M}) where {M} = Val{M}()
+@generated valsum(::Val{M}, ::Val{N}) where {M,N} = Val{M+N}()
+@inline valsum(::Val{M}, ::Val{N}, ::Val{K}, args...) where {M,N,K} = valsum(valsum(Val{M}(), Val{N}()), Val{K}(), args...)
+@inline valdims(::Any) = Val{1}()
+@inline valdims(::CartesianIndices{N}) where {N} = Val{N}()
+
+function append_loop_valdims!(valcall::Expr, loop::Loop)
+    if isstaticloop(loop)
+        push!(valcall.args, :(Val{1}()))
+    else
+        push!(valcall.args, Expr(:call, lv(:valdims), loop_boundary(loop)))
+    end
+    nothing
+end
+function subset_vptr!(ls::LoopSet, vptr::Symbol, indnum::Int, ind, previndices, loopindex)
     subsetvptr = Symbol(vptr, "_subset_$(indnum)_with_$(ind)##")
-    inde = ind isa Symbol ? Expr(:call, :-, ind, 1) : ind - 1
-    pushpreamble!(ls, Expr(:(=), subsetvptr, Expr(:call, lv(:subsetview), vptr, Expr(:call, Expr(:curly, :Val, indnum)), inde)))
+    valcall = Expr(:call, Expr(:curly, :Val, 1))
+    if indnum > 1
+        valcall = Expr(:call, lv(:valsum), valcall)
+        for i ∈ 1:indnum-1
+            if loopindex[i]
+                append_loop_valdims!(valcall, getloop(ls, previndices[i]))
+            else
+                for loopdep ∈ loopdependencies(ls.opdict[previndices[i]])
+                    append_loop_valdims!(valcall, getloop(ls, loopdep))
+                end
+            end
+        end
+    end
+    # @show valcall
+    indm1 = ind isa Integer ? ind - 1 : Expr(:call, :-, ind, 1)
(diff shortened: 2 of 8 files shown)


src/reconstruct_loopset.jl
```diff
@@ -378,7 +378,7 @@ function avx_loopset(instr, ops, arf, AM, LPSYM, LB, vargs)
     resize!(ls.loop_order, ls.loopsymbol_offsets[end])
     arraysymbolinds = gen_array_syminds(AM)
     opsymbols = [gensym(:op) for _ ∈ eachindex(ops)]
-    nopsv = calcnops.(Ref(ls), ops)
+    nopsv = [calcnops(ls, op) for op in ops]
```
timholy, line 381: Some time ago I found that compilation time was shorter for `map` than broadcasting, and I presume that extends to comprehensions. But this is purely cosmetic.

src/split_loops.jl
```diff
@@ -74,12 +74,12 @@ function lower_and_split_loops(ls::LoopSet)
             ls_2_lowered = if length(remaining_ops) > 1
                 lower_and_split_loops(ls_2)
             else
-                lower(ls_2, unrolled_2, tiled_2, vectorized_2, U_2, T_2)
+                lower(ls_2, order_2, unrolled_2, tiled_2, vectorized_2, U_2, T_2)
```
timholy, line 77: I think this and the next must have been a bug? If so, it's not caught by current tests. I am unsure how to go about designing one.

chriselrod: Yes, you're right, thanks for catching that.
Here is a testcase:
```julia
function twogemms!(Ab, Bb, Cb, A, B)
    M, N = size(C); K = size(B,1)
    @avx for m in 1:M, k in 1:K, n in 1:N
        Ab[m,k] += Cb[m,n] * B[k,n]
        Bb[k,n] += A[m,k] * Cb[m,n]
    end
end
```
This predictably throws a method error:
```julia
julia> M = 100;

julia> A = rand(M,M); B = rand(M,M); Cb = rand(M,M); Ab = similar(A); Bb = similar(B);

julia> twogemms!(Ab, Bb, Cb, A, B)
ERROR: MethodError: no method matching lower(::LoopVectorization.LoopSet, ::Symbol, ::Symbol, ::Symbol, ::Int64, ::Int64)
```
The motivation was that the autodiff code could produce loops like this. This particular example is what it should be producing as the reverse diff pass for `C = A * B` (`Cb`, `Ab`, and `Bb` being the partials with respect to `C`, `A`, and `B`).
It'd be more efficient to split it.
I should've added tests, also to make sure it really does split.

I shouldn't have merged until I actually added tests for the new functionality, even if all existing tests passed.

src/reconstruct_loopset.jl
chriselrod: Good call. I'll stick with comprehensions or maps.

chriselrod:
There is something else missing for the exponential filter demo:
```julia
function expfiltdim(x, dim::Integer, α)
    s = similar(x)
    Rpre = CartesianIndices(size(x)[1:dim-1])
    Rpost = CartesianIndices(size(x)[dim+1:end])
    _expfilt!(s, x, α, Rpre, size(x, dim), Rpost)
end

function _expfilt!(s, x, α, Rpre, n, Rpost)
    for Ipost in Rpost
        # Initialize the first value along the filtered dimension
        for Ipre in Rpre
            s[Ipre, 1, Ipost] = x[Ipre, 1, Ipost]
        end
        # Handle all other entries
        for i = 2:n
            for Ipre in Rpre
                s[Ipre, i, Ipost] = α*x[Ipre, i, Ipost] + (1-α)*s[Ipre, i-1, Ipost]
            end
        end
    end
    s
end
```
There is a loop-carried dependency:
```julia
s[Ipre, i, Ipost] = α*x[Ipre, i, Ipost] + (1-α)*s[Ipre, i-1, Ipost]
```
That is `s[Ipre, i, Ipost]` depends on the previous iteration, `s[Ipre, i - 1, Ipost]`. 

LoopVectorization doesn't yet support loop-carried dependencies, but it's something that would be very nice to support, since they show up in lots of code, and don't always prevent.

The simplest way to get them to work would be to prevent the loop with a loop-carried dependency (the `i` loop) from being vectorized and from being unrolled/tiled.
Disallowing vectorization is a fundamental limitation of SIMD. You can't vectorize `cumsum(::AbstractVector{<:Number}})`.
Disallowing unrolling is an implementation detail of LoopVectorization. It interleaves operations while lowering them.
```julia
Onα = 1 - α

s_1 = s[Ipre, i-1, Ipost]
s_2 = s[Ipre, i  , Ipost]
x_1 = x[Ipre, i, Ipost]
x_2 = x[Ipre, i+1, Ipost]
αx_1 = α * x_1
αx_2 = α * x_2
Onαs_1 = Onα * s_1
Onαs_2 = Onα * s_2
s[Ipre, i, Ipost] = αx_1 + Onαs_1
s[Ipre, i+1, Ipost] = αx_2 + Onαs_2
```
The idea behind this interleaving is to better support super-scalar parallelism. The operations next to each other wouldn't depend on each other, so no complicated reordering is needed by the CPU's scheduler to execute them in parallel.

The problem is that in this case, the above unrolling is wrong. `s_2` should equal `αx_1 + Onαs_1`, not `s[Ipre, i  , Ipost]`.
I think limiting unrolling these loops to `1` is an okay solution. One of the advantages of unrolling is to get that increased parallelism. But we would get a little less if this loop were lowered correctly:
```julia
Onα = 1 - α

s_1 = s[Ipre, i-1, Ipost]
x_1 = x[Ipre, i, Ipost]
Onαs_1 = Onα*s_1
αx_1 = α*x_1
s_2 = αx_1 + Onαs_1
s[Ipre, i  , Ipost] = s_2
x_2 = x[Ipre, i+1, Ipost]
αx_2 = α * x_2
Onαs_2 = Onα * s_2
s[Ipre, i+1, Ipost] = αx_2 + Onαs_2
```
`Onα * s_2` has to wait on `s_2 = αx_1 + Onαs_1` to complete before starting, but `αx_2` could still be computed in parallel with earlier operations.
The biggest advantage from unrolling this loop would be eliminating the load for `s_2`. With 2x unrolling, it needs just 3 loads, not 4. Because the loads are probably slower than the arithmetic (2x `*` + 2x `fma`), that's likely to still help performance.

chriselrod:
>  My presumption is that the easiest approach is to return `Δidxs` from calcnops and thus make `nopsv` a `Vector{Vector{Int}}`

Yes, I think that is the easiest approach.

An alternative that would be more complicated (at least in the short term) would be to define a `NopsDescription` struct to make a cleaner API for querying about the operations. That could make things simpler long term.
It could house `Δidxs` and `expandedv`.

Out of curiosity, what are your thoughts on explicit type signatures for all these functions? My thought was that using `nopsv::Vector{Int}` makes it easier to catch errors (including the method error on `lower` from the `loop_split` code), and ensures I look at each function taking `nopsv` as an argument.
A secondary thought is that I'd rather avoid compiling more than a single method of any of these functions.

timholy:
Thanks! Interesting about the loop-carried dependency, makes sense. Despite this example it is not an issue for most of ImageFiltering, only the `IIRGaussian` method I think. The most common uses of separable filters are FIR filtering.

> what are your thoughts on explicit type signatures for all these functions

Interesting question. In general if it's the case that you only expect particular input types, then adding checks essentially becomes a set of asserts that can catch bugs early. It's the type annotations that restrict what you can do with a function (e.g., `x::Vector{Float64}` when the method would work for any `AbstractArray`) that are annoying. I can't remember seen anything I'd disagree with.

I will go about making that change, and re-jiggering the test to avoid the loop-carried dependency. Maybe I'll put one in as a `@test_broken` just so you can reap the reward when it does become supported.

src/split_loops.jl
chriselrod: I implemented your fixes and added `twogemms!` to the unit tests.
I should probably add a `threegemms!` as well to hit the last remaining un-evaluated line in `split_loops!`. https://github.com/chriselrod/LoopVectorization.jl/blob/0e0c418f4fa9a976c09a0fb2b9cd7505632c3e16/src/split_loops.jl#L75
Thanks again!

test/miscellaneous.jl
```diff
@@ -652,4 +656,54 @@
...
+            s
+        end
+        function smoothdim_avx!(s, x, α, Rpre, irng::AbstractUnitRange, Rpost)
+            ifirst, ilast = first(irng), last(irng)
+            ifirst > ilast && return s
+            @avx for Ipost in Rpost
+                # Initialize the first value along the filtered dimension
+                for Ipre in Rpre
+                    s[Ipre, ifirst, Ipost] = x[Ipre, ifirst, Ipost]
+                end
```
timholy, line 687: At least for `d=1` in the test-loop at the bottom of this file, the test failure is due to this loop failing to run for all of `Rpost`; it only runs for `first(Rpost)`. (And I do mean `Rpost` rather than `Rpre`.) I am guessing this is a loop-ordering thing? I looked at the expanded code but struggled to figure out which pieces come from what parts of this code; if you get a chance, mind looking at that? For the 3d test below, test failures occur for `d=1` and `d=3`, whereas it passes on `d=2`. Would probably be best to reenable the 5d (or at least, 4d) case before merging, since you need `d=5` to test the case of indexing with `x[::CI{2}, ::Int, ::CI{2}]` (`CI=CartesianIndex`).

chriselrod: I think heterogenous `nops` should either only be allowed for `memload` and `memstore` operations, or perhaps also allowed for `compute` iff `Set(Δidxs)` == `Set([1,n])` (i.e., broadcasting some `nops == 1`s to match the others).

I'd lean toward the former now, for simplicity. That means, maintain the old behavior, except for indexing operations.
That means you could make the `@assert` immediate, but conditional on `!accesses_memory(op)`.

I'll take a deeper look at the code in the next few hours and leave more comments.

chriselrod: I found a couple problems. I'll push when I've fixed them and leave a few comments.

src/memory_ops_common.jl
```diff
@@ -29,17 +29,46 @@ function add_vptr!(ls::LoopSet, array::Symbol, vptrarray::Symbol = vptr(array),
...
+
+function append_loop_valdims!(valcall::Expr, loop::Loop)
+    if isstaticloop(loop)
+        push!(valcall.args, :(Val{1}()))
+    else
+        push!(valcall.args, Expr(:call, lv(:valdims), loop_boundary(loop)))
+    end
+    nothing
+end
+function subset_vptr!(ls::LoopSet, vptr::Symbol, indnum::Int, ind, previndices, loopindex)
```
chriselrod, line 48: For now, I decided it would be easier to update the old `subsetview` function. This function is used for constant slices, like by `ifirst`. This function gets called very early, to avoid storing or representing that there was an index at all.
This probably caused the failures you observed, because it assumed the preceding indices corresponded to one axis each. That assumption was violated when `d == 1` and `d == 3`, where the first index corresponded to 0 and 2 axes, instead.

It now calls `valdims` on each of the preceding loops instead.
I currently defined `valdims` to return `Val(1)` for everything except `CartesianIndices{N}`, where it returns `Val(N)` instead.

src/condense_loopset.jl
```diff
@@ -327,7 +337,8 @@ function setup_call_noinline(ls::LoopSet, U = zero(Int8), T = zero(Int8))
         push!(q.args, ex)
     end
     insert!(call.args, 8, Expr(:call, Expr(:curly, :Val, vptrarrays)))
-    insert!(call.args, 9, Expr(:call, Expr(:curly, :Val, vptrsubsetdims)))
+    # insert!(call.args, 9, Expr(:call, Expr(:curly, :Val, vptrsubsetdims)))
+    insert!(call.args, 9, vptrsubsetdims)
```
chriselrod, line 341: This function parsed `Val{N}()` expressions to extract the `N`, and create a expression for `Val{tuple_of_Ns}()`. This broke as soon as the `Val{N}()` expressions were replaced with calls that returned `Val`s.
So now, instead of parsing expressions, it makes them arguments to a function that will return the desired answer.

The purpose of that song and dance is so that `__avx__!` is not inlined, and calls `_avx_!`, which is inlined.
Within the `__avx__!` call, it'll subset the arrays, get pointers, etc, so that LLVM still has this aliasing / relative pointer offset information available. This reduced the performance disparity between setting `inline=false` and `inline=true`. The latter is the default.
This is not a good solution, but a hack added on later when I decided to add this.

test/miscellaneous.jl
chriselrod: The `tile=(1,1)` is unnecessary. I just added that while experimenting.
But I think `@avx` is being overly aggressive in unrolling this. I'll run benchmarks later.

test/miscellaneous.jl
```diff
@@ -680,30 +680,46 @@ using Test
...
                 # Handle all other entries
-                for i = ifirst+1:ilast
+                for i = ifirst:ilast
                     for Ipre in Rpre
-                        s[Ipre, i, Ipost] = α*x[Ipre, i, Ipost] + (1-α)*x[Ipre, i-1, Ipost]
+                        xi = x[Ipre, i, Ipost]
+                        xim = ifelse(i == ifirst, xi, x[Ipre, i-1, Ipost])
+                        s[Ipre, i, Ipost] = α*xi + (1-α)*xim
                     end
                 end
```
chriselrod, line 705: Not as pretty, but another way to write the function. 

I think this may risk crashing because of the read to `x[Ipre, ifirst-1, Ipost]`, even though that value wont be used, so I'm not inclined to recommend this until I add explicit support for conditional loads, like the conditional store support it has now.

test/miscellaneous.jl
```diff
@@ -680,30 +680,46 @@ using Test
...
-        dest1, dest2 = similar(x), similar(x)
+        x = rand(11,11,11,11,11);
+        dest1, dest2 = similar(x), similar(x);
         α = 0.3
         for d = 1:ndims(x)
-            Rpre  = CartesianIndices(axes(x)[1:d-1])
-            Rpost = CartesianIndices(axes(x)[d+1:end])
+            # @show d
+            Rpre  = CartesianIndices(axes(x)[1:d-1]);
+            Rpost = CartesianIndices(axes(x)[d+1:end]);
```
chriselrod, line 716: The semicolons are because I don't want to be bombarded by output when running these interactively.

src/reconstruct_loopset.jl
```diff
@@ -239,11 +239,12 @@ function calcnops(ls::LoopSet, os::OperationStruct)
     offsets = ls.loopsymbol_offsets
     idxs = loopindex(ls, os.loopdeps, 0x04)  # FIXME DRY
     iszero(length(idxs)) && return 1
-    return map(i->offsets[i+1]-offsets[i], idxs)
+    return maximum(i->offsets[i+1]-offsets[i], idxs)
 end
```
chriselrod, line 237: For now, only loads and stores are allowed to have a mix of `nops` as inputs.
`nops` for loads and stores are also always `1`, so we don't need a vector.

chriselrod:
Tests passed locally (I re-enabled 5d). I set Travis to allow failures on master, because master currently seems to be broken.

I made a few comments discussing some of the changes.

chriselrod:
5-d tests pass on 1.1 and 1.4, and a local several-day-old master.

test/miscellaneous.jl
chriselrod: This did cause crashes on Travis. I'll work on turning this into a conditional load.

test/miscellaneous.jl
timholy: But it's nice that I don't have to think about cache-friendliness when writing out those loops!

timholy:
Wow, amazing work! Sorry I didn't bring this closer to completion, I was hoping it was a simple matter.

Certainly I'd approve merging this whenever you are satisfied with it. Anything else I can do to help?

chriselrod:
Thanks, I appreciate your contributions and getting this started! I'd also be happy to write up how different things work. Some of that could also help me figure out what it should look like after components get rewritten/refactored with internal APIs.

I merged because tests passed. Performance needs work, but I can handle that later.

Performance work mostly means adding support for recognizing constant integer offsets, and then using these in `determinestrategy.jl` to encourage it to unroll the loops that allow eliminating loads (i.e., the `i` loop here).
You're also more than welcome to take a stab at it. That would also lay the same groundwork for recognized loop-carried dependencies.

> I looked at the expanded code but struggled to figure out which pieces come from what parts of this code

Anything in particular I can answer?

timholy:
> Anything in particular I can answer?

I need to go back and figure out how operations work in greater detail. https://chriselrod.github.io/LoopVectorization.jl/latest/devdocs/loopset_structure/ was really helpful in getting going, but now I see I should go back to that page and really internalize all the stuff there. So mostly I think this is me just not doing my homework carefully enough.

Some things that might help:

- comments or a docstring associated with each element of [`OperationType`](https://github.com/chriselrod/LoopVectorization.jl/blob/1caaa110dcdb419adf7728d66761082f9ea4477b/src/operations.jl#L59-L65)
- comments on the fields of [`Operation`](https://github.com/chriselrod/LoopVectorization.jl/blob/1caaa110dcdb419adf7728d66761082f9ea4477b/src/operations.jl#L70-L103)

I can take a stab at some of these, it would be good for me.

Issue #73, opened 2020-03-13
Allow unrolling an arbitrary number of loops (or at least up to 4) and improve load elimination


chriselrod:
I've simplified the lowering code as a first step, but it still expects input in terms of unrolled and tiled. This should be replaced with a vector of integers indicating by how much each loop is unrolled (1 means not unrolled).

The more complicated step will be generalizing the constrained optimization problem within [solve_tilesize](https://github.com/chriselrod/LoopVectorization.jl/blob/master/src/determinestrategy.jl#L247) to an arbitrary number of dimensions.

I think I'll have it handle unrolling a single loop as a special case. The `solve_tilesize` function tries to occupy all the registers, which wouldn't really be appropriate for a single loop.

It also needs the ability to recognize more kinds of loads that can be eliminated. In particular, it should be able to realize that if an array is accessed with both `A[i]` and `A[i+1]`, unrolling `i` will only require a single extra load, not 2 extra loads, because `A[i+1]` == `A[(i+1)]`.

The motivating example was inspired by @timholy , although before LoopVectorization is going to see major benefits over the current performance, it will have to be able to realize t

Most of the code here is also in the [offset array tests](https://github.com/chriselrod/LoopVectorization.jl/blob/master/test/offsetarrays.jl).
```julia
using LoopVectorization, OffsetArrays, BenchmarkTools, Test
function old2d!(out::AbstractMatrix, A::AbstractMatrix, kern)
    rng1k, rng2k = axes(kern)
    rng1,  rng2  = axes(out)
    for j in rng2, i in rng1
        tmp = zero(eltype(out))
        for jk in rng2k, ik in rng1k
            tmp += A[i+ik,j+jk]*kern[ik,jk]
        end
        out[i,j] = tmp
    end
    out
end
function avx2douter!(out::AbstractMatrix, A::AbstractMatrix, kern)
    rng1k, rng2k = axes(kern)
    rng1,  rng2  = axes(out)
    @avx for j in rng2, i in rng1
        tmp = zero(eltype(out))
        for jk in rng2k, ik in rng1k
            tmp += A[i+ik,j+jk]*kern[ik,jk]
        end
        out[i,j] = tmp
    end
    out
end

T = Float32
A = rand(T, 100, 100);
kern = OffsetArray(rand(T, 3, 3), -1:1, -1:1);
out1 = OffsetArray(similar(A, size(A).-2), 1, 1);   # stay away from the edges of A
[...]

PR #94, opened 2020-04-11
Add more documentation on the internals

timholy:
While most of this should hopefully be OK, notice that I've embedded a number of questions where I am uncertain of the answer. It would be best to resolve or delete these questions before merging. @chriselrod, you are authorized to just edit the files or answer here and I will make changes.

One oddity: Documenter complains about not being able to find the docstring for `LoopVectorization.vfilter!` and then later complains that we didn't put it in a `@docs` block. I am confused by this.

Hope the additional dependency on DocStringExtensions is OK. If not, we can delete it.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -90,6 +90,29 @@ julia> b ≈ c
 true
 ```
 
+# Extended help
+
+Advanced users can customize the implementation of the `@avx`-annotated block
+using keyword arguments:
+
+```
+@avx inline=false unroll=2 body
+```
+
+where `body` is the code of the block (e.g., `for ... end`).
+
+`inline` is a Boolean. When `true` (the default), `body` will be directly inlined
+into the function (via a forced-inlining call to `_avx_!`).
+When `false`, it will call `__avx__!` instead, letting Julia's own inlining engine
+determine whether the call to `__avx__!` should be inlined. (Typically, it won't.)
+In priniciple, first calling `__avx__!` (which itself calls `_avx_!`) can sometimes
+allow better code generation.
+One can find some circumstances where `inline=true` is faster, and other circumstances
+where `inline=false` is faster, so the best setting may require experimentation.
+
+`unroll` is an integer that specifies the loop unrolling factor, or a
+tuple `(4, 2)` signaling that the generated code should unroll more than
+one loop.
 """
 macro avx(q)
     q = macroexpand(__module__, q)
@@ -121,7 +144,7 @@ function check_unroll(arg)
             T = convert(Int8, tup.args[2])
         else
             return nothing
-        end    
+        end
     else
         return nothing
     end
@@ -179,4 +202,3 @@ macro avx_debug(q)
     q = macroexpand(__module__, q)
     esc(LoopVectorization.setup_call_debug(LoopSet(q, __module__)))
 end
-
--- a/src/costs.jl
+++ b/src/costs.jl
@@ -6,7 +6,12 @@ else
 end
 
 
+"""
+    Instruction
 
+`Instruction` represents a function via its module and symbol. It is
+similar to a `GlobalRef` and may someday be replaced by `GlobalRef`.
+"""
 struct Instruction
     mod::Symbol
     instr::Symbol
@@ -36,10 +41,24 @@ Base.isequal(ins1::Instruction, ins2::Instruction) = (ins1.instr === ins2.instr)
 
 const LOOPCONSTANT = Instruction(Symbol("LOOPCONSTANTINSTRUCTION"))
 
+"""
+    InstructionCost
+
+Store parameters related to performance for individual CPU instructions.
+
+$(TYPEDFIELDS)
+"""
 struct InstructionCost
+    "A flag indicating how instruction cost scales with vector width (128, 256, or 512 bits)"
     scaling::Float64 # sentinel values: -3 == no scaling; -2 == offset_scaling, -1 == linear scaling, >0 ->  == latency == reciprocal throughput
+    """The number of clock cycles per operation when many of the same operation are repeated in sequence.
+    Think of it as the inverse of the flow rate at steady-state. It is typically ≤ the `scalar_latency`."""
     scalar_reciprocal_throughput::Float64
+    """The minimum delay, in clock cycles, associated with the instruction.
+    Think of it as the delay from turning on a faucet to when water starts coming out the end of the pipe.
+    See also `scalar_reciprocal_throughput`."""
     scalar_latency::Int
+    "Number of floating-point registered used"
     register_pressure::Int
 end
 InstructionCost(sl::Int, srt::Float64, scaling::Float64 = -3.0) = InstructionCost(scaling, srt, sl, 0)
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -1,7 +1,21 @@
+"""
+    ArrayReference
 
+A type for encoding an array reference `A[i,j]` occurring inside an `@avx` block.
+
+# Fields
+
+$(TYPEDFIELDS)
+"""
 struct ArrayReference
+    "The array variable"
     array::Symbol
+    "The list of indices (e.g., `[:i, :j]`), or `name(op)` for computed indices."
     indices::Vector{Symbol}
+    """Index offset, e.g., `a[i+7]` would store the `7`. `offsets` is also used
+    to help identify opportunities for avoiding reloads, for example in `y[i] = x[i] - x[i-1]`,
+    the previous load `x[i-1]` can be "carried over" to the next iteration.
+    Only used for small (`Int8`) offsets."""    
     offsets::Vector{Int8}
 end
 ArrayReference(array, indices) = ArrayReference(array, indices, zeros(Int8, length(indices)))
@@ -26,9 +40,21 @@ function Base.isequal(x::ArrayReference, y::ArrayReference)
     end
     true
 end
+"""
+    ArrayReferenceMeta
+
+A type similar to [`ArrayReference`](@ref) but holding additional information.
+
+# Fields
+
+$(TYPEDFIELDS)
+"""
 struct ArrayReferenceMeta
+    "The `ArrayReference`"
     ref::ArrayReference
+    "A vector of Bools indicating whether each index is a loop variable (`false` for operation-computed indices)"
     loopedindex::Vector{Bool}
+    "Variable holding the pointer to the array's underlying storage"
     ptr::Symbol
 end
 function ArrayReferenceMeta(ref::ArrayReference, loopedindex, ptr = vptr(ref))
@@ -56,6 +82,10 @@ Base.:(==)(x::ArrayReferenceMeta, y) = false
 
 abstract type AbstractLoopOperation end
 
+"""
+`OperationType` is an `@enum` for classifying supported operations that can appear in
+`@avx` blocks. Type `LoopVectorization.OperationType` to see the different types.
+"""
 @enum OperationType begin
     constant
     memload
@@ -63,22 +93,78 @@ abstract type AbstractLoopOperation end
     memstore
     loopvalue
 end
+"An operation setting a variable to a constant value (e.g., `a = 0.0`)" constant
+"An operation setting a variable from a memory location (e.g., `a = A[i,j]`)" memload
+"An operation computing a new value from one or more variables (e.g., `a = b + c`)" compute
+"An operation storing a value to a memory location (e.g., `A[i,j] = a`)" memstore
+"""
+`loopvalue` indicates an loop variable (`i` in `for i in ...`). These are the "parents" of `compute`
+operations that involve the loop variables.
+"""
+loopvalue
 
 # TODO: can some computations be cached in the operations?
 """
+    Operation
+
+A structure to encode a particular action occuring inside an `@avx` block.
+
+# Fields
+
+$(TYPEDFIELDS)
+
+# Example
+
+```jldoctest Operation; filter = r"\\"##.*\\""
+julia> using LoopVectorization
+
+julia> AmulBq = :(for m ∈ 1:M, n ∈ 1:N
+           C[m,n] = zero(eltype(B))
+           for k ∈ 1:K
+               C[m,n] += A[m,k] * B[k,n]
+           end
+       end);
+
+julia> lsAmulB = LoopVectorization.LoopSet(AmulBq);
+
+julia> LoopVectorization.operations(lsAmulB)
+6-element Array{LoopVectorization.Operation,1}:
+ var"##RHS#253" = var"##zero#254"
+ C[m, n] = var"##RHS#253"
+ var"##tempload#255" = A[m, k]
+ var"##tempload#256" = B[k, n]
+ var"##RHS#253" = LoopVectorization.vfmadd_fast(var"##tempload#255", var"##tempload#256", var"##RHS#253")
+ var"##RHS#253" = LoopVectorization.identity(var"##RHS#253")
+```
+Each one of these lines is a pretty-printed `Operation`.
 """
 mutable struct Operation <: AbstractLoopOperation
+    """A unique identifier for this operation.
+    `identifer(op::Operation)` returns the index of this operation within `operations(ls::LoopSet)`."""
     identifier::Int
+    """The name of the variable storing the result of this operation.
+    For `a = val` this would be `:a`. For array assignments `A[i,j] = val` this would be `:A`."""
     variable::Symbol
+    "Intended to be the size of the result, in bytes. Often inaccurate, not to be relied on."
     elementbytes::Int
+    "The specific operator, e.g., `identity` or `+`"
     instruction::Instruction
+    "The [`OperationType`](@ref) associated with this operation"
     node_type::OperationType
+    "The loop variables this operation depends on"
     dependencies::Vector{Symbol}
+    "Additional loop dependencies that must execute before this operation can be performed successfully (often needed in reductions)"
     reduced_deps::Vector{Symbol}
+    "Operations whose result this operation depends on"
     parents::Vector{Operation}
+    "For `memload` or `memstore`, encodes the array location"
     ref::ArrayReferenceMeta
+    "`gensymmed` name of result."
     mangledvariable::Symbol
+    """Loop variables that *consumers* of this operation depend on.
+    Often used in reductions to replicate assignment of initializers when unrolling."""
     reduced_children::Vector{Symbol}
+
     function Operation(
         identifier::Int,
         variable,
@@ -129,7 +215,7 @@ const NOPARENTS = Operation[]
 function Base.show(io::IO, op::Operation)
     if isconstant(op)
         if op.instruction === LOOPCONSTANT
-            
+
             print(io, Expr(:(=), op.variable, 0))
         else
             print(io, Expr(:(=), op.variable, op.instruction.instr))
--- a/src/reconstruct_loopset.jl
+++ b/src/reconstruct_loopset.jl
@@ -417,8 +417,26 @@ function _avx_loopset(OPSsv, ARFsv, AMsv, LPSYMsv, LBsv, vargs)
         AMsv, LPSYMsv, LBsv, vargs
     )
 end
+const _body_ = Ref{Any}(nothing)
+"""
+    _avx_!(ut, ops, arf, am, lpsym, lb, vargs...)
+
+Execute an `@avx` block. The block's code is represented via the arguments:
+- `ut` is `Val((U,T))`, where `U` is the unrolling factor and `T` ?has something to do with tiling?
+- `ops` is `Tuple{mod1, sym1, op1, mod2, sym2, op2...}` encoding the operations of the loop.
+  `mod` and `sym` encode the module and symbol of the called function; `op` is an [`OperationStruct`](@ref)
+  encoding the details of the operation.
+- `arf` is `Tuple{arf1, arf2...}`, where each `arfi` is an [`ArrayRefStruct`](@ref) encoding
+  an array reference.
+- `am` contains miscellaneous data about the LoopSet (see `process_metadata!`)
+- `lpsym` is `Tuple{:i,:j,...}`, a Tuple of the "loop symbols", i.e. the item variable `i` in `for i ∈ iter`
+- `lb` is `Tuple{RngTypei,RngTypej,...}`, a Tuple encoding syntactically-knowable information about
+  the iterators corresponding to `lpsym`. For example, in `for i ∈ 1:n`, the `1:n` would be encoded with
+  `StaticLowerUnitRange(1)` because the lower bound of the iterator can be determined to be 1.
+- `vargs...` holds the encoded pointers of all the arrays (see `VectorizationBase`'s various pointer types).
+"""
 @generated function _avx_!(::Val{UT}, ::Type{OPS}, ::Type{ARF}, ::Type{AM}, ::Type{LPSYM}, lb::LB, vargs...) where {UT, OPS, ARF, AM, LPSYM, LB}
     1 + 1 # Irrelevant line you can comment out/in to force recompilation...
     ls = _avx_loopset(OPS.parameters, ARF.parameters, AM.parameters, LPSYM.parameters, LB.parameters, vargs)
-    avx_body(ls, UT)
+    return _body_[] = copy(avx_body(ls, UT))
 end
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Chris Elrod <elrodc@gmail.com>"]
 version = "0.7.0"
 
 [deps]
+DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 SIMDPirates = "21efa798-c60a-11e8-04d3-e1a92915a26a"
@@ -12,6 +13,7 @@ UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 VectorizationBase = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
 
 [compat]
+DocStringExtensions = "0.8"
 OffsetArrays = "1"
 SIMDPirates = "0.7.13"
 SLEEFPirates = "0.4.4"
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -21,7 +21,8 @@ makedocs(;
             "devdocs/loopset_structure.md",
             "devdocs/constructing_loopsets.md",
             "devdocs/evaluating_loops.md",
-            "devdocs/lowering.md"
+            "devdocs/lowering.md",
+            "devdocs/reference.md"
         ]
     ],
     # repo="https://github.com/chriselrod/LoopVectorization.jl/blob/{commit}{path}#L{line}",
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -22,5 +22,5 @@ vmapntt!
 
 ```@docs
 vfilter
-vfilter!
+LoopVectorization.vfilter!
 ```
--- a/docs/src/devdocs/reference.md
+++ b/docs/src/devdocs/reference.md
@@ -1,0 +1,42 @@
+# Internals reference
+
+## Operation types
+
+```@docs
+LoopVectorization.OperationType
+LoopVectorization.constant
+LoopVectorization.memload
+LoopVectorization.compute
+LoopVectorization.memstore
+LoopVectorization.loopvalue
+```
+
+## Operation
+
+```@docs
+LoopVectorization.Operation
+```
+
+## Instructions and costs
+
+```@docs
+LoopVectorization.Instruction
+LoopVectorization.InstructionCost
+```
+
+## Array references
+
+```@docs
+LoopVectorization.ArrayReference
+LoopVectorization.ArrayReferenceMeta
+```
+
+## Condensed types
+
+These are used when encoding the `@avx` block as a type parameter for passing through
+to the `@generated` function.
+
+```@docs
+LoopVectorization.ArrayRefStruct
+LoopVectorization.OperationStruct
+```
--- a/src/LoopVectorization.jl
+++ b/src/LoopVectorization.jl
@@ -16,6 +16,7 @@ using SLEEFPirates: pow
 using Base.Broadcast: Broadcasted, DefaultArrayStyle
 using LinearAlgebra: Adjoint, Transpose
 using Base.Meta: isexpr
+using DocStringExtensions
 
 
 const SUPPORTED_TYPES = Union{Float16,Float32,Float64,Integer}
--- a/src/condense_loopset.jl
+++ b/src/condense_loopset.jl
@@ -5,7 +5,9 @@ Base.:|(u::Unsigned, it::IndexType) = u | UInt8(it)
 Base.:(==)(u::Unsigned, it::IndexType) = (u % UInt8) == UInt8(it)
 
 """
-`ArrayRefStruct` stores a representation of an array-reference expression such as `A[i,j]`.
+    ArrayRefStruct
+
+A condensed representation of an [`ArrayReference`](@ref).
 It supports array-references with up to 8 indexes, where the data for each consecutive index is packed into corresponding 8-bit fields
 of `index_types` (storing the enum `IndexType`), `indices` (the `id` for each index symbol), and `offsets` (currently unused).
 """
@@ -53,6 +55,11 @@ function ArrayRefStruct(ls::LoopSet, mref::ArrayReferenceMeta, arraysymbolinds::
     ArrayRefStruct{mref.ref.array,mref.ptr}( index_types, indices, offsets )
 end
 
+"""
+    OperationStruct
+
+A condensed representation of an [`Operation`](@ref).
+"""
 struct OperationStruct <: AbstractLoopOperation
     # instruction::Instruction
     loopdeps::UInt64
--- a/test/offsetarrays.jl
+++ b/test/offsetarrays.jl
@@ -180,7 +180,7 @@ T = Float64
            out[I] = tmp
        end
        out
-   end
+    end
     function avxgeneric2!(out, A, kern)
        @avx for I in CartesianIndices(out)
            tmp = zero(eltype(out))
@@ -225,7 +225,7 @@ T = Float64
         @test avxgeneric!(out4, A, kern) ≈ out1
         fill!(out4, NaN);
         @test avxgeneric!(out4, A, skern) ≈ out1
-        
+
         fill!(out4, NaN); @test avxgeneric2!(out4, A, kern) ≈ out1
         fill!(out4, NaN); @test avxgeneric2!(out4, A, skern) ≈ out1
     end


src/costs.jl
```diff
@@ -6,7 +6,15 @@ else
...
 
 
+"""
+    Instruction
 
+`Instruction` is like a `GlobalRef` except it isn't. (Is this because you need it to be
+encodable in a type parameter? How do you handle submodules? Would making `mod` the result
+of `fullname` fix that?)
+
+Also, are these just low-level CPU instructions or is `Base.sin` allowed?
```
chriselrod, line 16: `sin` is included in the `COST` dict. `Base.sin(::SVec)` dispatches to a SIMD implementation.
If the user is on Linux with a recent version of GLIBC, it will use the method in `libmvec.so.1`. Otherwise, it'll use the implementation defined in `SLEEFPirates.jl`.

These represent function calls, which can be expensive, like [clenshaw](https://github.com/chriselrod/LoopVectorization.jl/blob/master/test/runtests.jl#L5) called in the [clenshaw!](https://github.com/chriselrod/LoopVectorization.jl/blob/master/test/miscellaneous.jl#L205) test.

src/costs.jl
```diff
@@ -6,7 +6,15 @@ else
 end
 
 
+"""
+    Instruction
 
+`Instruction` is like a `GlobalRef` except it isn't. (Is this because you need it to be
+encodable in a type parameter? How do you handle submodules? Would making `mod` the result
+of `fullname` fix that?)
```
chriselrod, line 14: Good suggestion. We could probably replace it with `GlobalRef`s.
We can't encode `Instruction` in a type parameter either, on a couple day old Julia master: 
```julia
julia> Val((:a, :b))
Val{(:a, :b)}()

julia> using LoopVectorization: Instruction

julia> Val(Instruction(:foo))
ERROR: TypeError: in Type, in parameter, expected Type, got a value of type Instruction
Stacktrace:
 [1] Val(::Instruction) at ./essentials.jl:696
 [2] top-level scope at REPL[13]:1
```
(condense/reconstruct)_loopset.jl use a vector of symbols.

I think this entire part of the code can be changed.
If you look at the `Base.Expr(::Instruction, args...)` method defined a few lines lower, you'll notice it more or less ignores the module; it'll only using a `Expr(:(.), mod, f)`/`GlobalRef` if the module is `LoopVectorization`.

This is because it's (now) supposed to work through one of the following means:
1) LoopVectorization "recognizes" the function, and it's defined in the LoopVectorization module's scope. Then it'll lower the function `f` as `LoopVectorization.f`
2) It will pass the function as an argument to `_avx_!`, and there will be a `gensym`ed name. Because the function has to be called within `_avx_!` within LoopVectorization, this is the best solution I'm aware of to make it inferrable and avoid world-age problems.

That wasn't the original design. `Instruction` could probably be replaced with a `Symbol`.

src/operations.jl
```diff
@@ -1,7 +1,18 @@
...
+
+# Fields
 
+$(TYPEDFIELDS)
+"""
 struct ArrayReference
+    "The array variable"
     array::Symbol
+    "The list of indices (e.g., `[:i, :j]`). Not sure what this does for computed indices, maybe that chains with Operations."
     indices::Vector{Symbol}
```
chriselrod, line 14: This should probably be combined with `ArrayReferenceMeta`, because it doesn't provide enough information by itself.
 
For computed indices, it will either
1. If the index is computed as a constant offset representable by an `Int8`, it'll set the corresponding index in `offsets` to that offset.
2. Define an `Operation` performing the computation. `name(op)` will be the corresponding index of `indices`.

src/operations.jl
```diff
@@ -1,7 +1,18 @@
...
+    "Unused?"
     offsets::Vector{Int8}
```
chriselrod, line 16: It's now used. It indicates a constant offset.
If it is not 0, it'll add the offset to the index while lowering.

src/operations.jl
```diff
@@ -26,9 +37,21 @@ function Base.isequal(x::ArrayReference, y::ArrayReference)
...
+
+# Fields
+
+$(TYPEDFIELDS)
+"""
 struct ArrayReferenceMeta
+    "The `ArrayReference`"
     ref::ArrayReference
+    "A vector of Bools indicating whether each index is a loop variable"
     loopedindex::Vector{Bool}
```
chriselrod, line 53: If an index is computed (and not representable by an `offset`, e.g. if it's a function of two loop variables), an operation will be added, and the corresponding `loopedindex` set to `false`.

src/operations.jl
```diff
@@ -56,29 +79,83 @@ Base.:(==)(x::ArrayReferenceMeta, y) = false
...
     memload
     compute
     memstore
     loopvalue
 end
+"An operation setting a variable to a constant value (e.g., `a = 0.0`)" constant
+"An operation setting a variable from a memory location (e.g., `a = A[i,j]`)" memload
+"An operation computing a new value from one or more variables (e.g., `a = b + c`)" compute
+"An operation storing a value to a memory location (e.g., `A[i,j] = a`)" memstore
+"??" loopvalue
```
chriselrod, line 97:
```julia
for i in 1:4, j in 0:3
    s += i * a[ i - j ]
end
```
In the operations
`i - j` and `i * ...`, `i` and `j` are represented as `loopvalue`s. LoopValues aren't needed for indexing, but are whenever they're a parent of a `compute` operation.

src/operations.jl
```diff
@@ -56,29 +79,83 @@ Base.:(==)(x::ArrayReferenceMeta, y) = false
...
+ var"##tempload#255" = A[m, k]
+ var"##tempload#256" = B[k, n]
+ var"##RHS#253" = LoopVectorization.vfmadd_fast(var"##tempload#255", var"##tempload#256", var"##RHS#253")
+ var"##RHS#253" = LoopVectorization.identity(var"##RHS#253")
+```
+Each one of these lines is a pretty-printed `Operation`.
 """
 mutable struct Operation <: AbstractLoopOperation
+    "A unique number assigned to each operation"
     identifier::Int
```
chriselrod, line 136: Importantly, it corresponds to 1 less than its index in `operations(ls)`.
The function `identifier(op::Operation)` returns `op.identifier + 1`.

I liked the idea of defining the identifier with `length(operations(ls))`. I don't know where the best place for the `+1` is.
Inside the `identifier` function at least doesn't require repeating the code `+ 1`. But it'll get called frequently (not a big deal), and it should be clear that `identifier` should be accessed through the `identifier` function, rather than `getproperty`.

src/operations.jl
```diff
@@ -56,29 +79,83 @@ Base.:(==)(x::ArrayReferenceMeta, y) = false
...
+    """The name of the variable storing the result of this operation.
+    For `a = val` this would be `:a`. For array assignments `A[i,j] = val` this would be `:A`."""
     variable::Symbol
+    "The size of the result, in bytes. This is not reliable prior to loop reconstruction when types are known."
     elementbytes::Int
```
chriselrod, line 141: Not reliable after either. Needs work.

src/operations.jl
```diff
@@ -56,29 +79,83 @@ Base.:(==)(x::ArrayReferenceMeta, y) = false
...
+    "The specific operator, e.g., `identity` or `+`"
     instruction::Instruction
+    "The [`OperationType`](@ref) associated with this operation"
     node_type::OperationType
+    "The loop variables this operation depends on"
     dependencies::Vector{Symbol}
+    "??"
     reduced_deps::Vector{Symbol}
```
chriselrod, line 149: Repeating the `Operation` example:
```julia
julia> using LoopVectorization

julia> AmulBq = :(for m ∈ 1:M, n ∈ 1:N
           C[m,n] = zero(eltype(B))
           for k ∈ 1:K
               C[m,n] += A[m,k] * B[k,n]
           end
       end);

julia> lsAmulB = LoopVectorization.LoopSet(AmulBq);

julia> LoopVectorization.operations(lsAmulB)
6-element Array{LoopVectorization.Operation,1}:
 var"##RHS#265" = var"##zero#266"
 C[m, n] = var"##RHS#265"
 var"##tempload#267" = A[m, k]
 var"##tempload#268" = B[k, n]
 var"##RHS#265" = LoopVectorization.vfmadd_fast(var"##tempload#267", var"##tempload#268", var"##RHS#265")
 var"##RHS#265" = LoopVectorization.identity(var"##RHS#265")

julia> LoopVectorization.loopdependencies.(LoopVectorization.operations(lsAmulB))
6-element Array{Array{Symbol,1},1}:
 [:m, :n]
 [:m, :n]
 [:m, :k]
 [:k, :n]
 [:m, :k, :n]
 [:m, :n]

julia> LoopVectorization.reduceddependencies.(LoopVectorization.operations(lsAmulB))
6-element Array{Array{Symbol,1},1}:
 0-element Array{Symbol,1}
 [:k]
 0-element Array{Symbol,1}
 0-element Array{Symbol,1}
 [:k]
 [:k]
```
It says that operation depends on a reduction over that loop. Here it means that 
`C[m,n] = RHS265` 
and
`var"##RHS#265" = LoopVectorization.identity(var"##RHS#265")`
depend on `m` and `n` directly -- so that they need to be placed inside the `m` and `n` loops -- but that they also depend on a reduction over the `k` loop, and thus have to be placed after the loop.
`var"##RHS#265" = var"##zero#266"`, on the other hand, would be placed before the `k` loop, as it does not depend on it.

`var"##RHS#265" = LoopVectorization.vfmadd_fast(var"##tempload#267", var"##tempload#268", var"##RHS#265")` depends on all three loops, but is a reduction across the `k` loop. I can't immediately recall where or if it matters when `k in loopdependencies(op)`, but it is important for the first two example ops that do not depend directly on `k`.

src/operations.jl
```diff
@@ -56,29 +79,83 @@ Base.:(==)(x::ArrayReferenceMeta, y) = false
...
+    "Operations whose result this operation depends on"
     parents::Vector{Operation}
+    "For `memload` or `memstore`, encodes the array location"
     ref::ArrayReferenceMeta
+    "??"
     mangledvariable::Symbol
+    "??"
     reduced_children::Vector{Symbol}
```
chriselrod, line 157: Continuing from `reduced_deps`
```julia
julia> LoopVectorization.reducedchildren.(LoopVectorization.operations(lsAmulB))
6-element Array{Array{Symbol,1},1}:
 [:k]
 0-element Array{Symbol,1}
 0-element Array{Symbol,1}
 0-element Array{Symbol,1}
 0-element Array{Symbol,1}
 0-element Array{Symbol,1}
```
The operation `var"##RHS#265" = var"##zero#266"` does not depend on the `k` loop, and its parents don't depend on it, but its children do.
If the `k` look is vectorized `var"##RHS#265"` will have to be a vector.
If the `k` look is unrolled by `U`, we'll need `U` copies of `var"##RHS#265"`.

Similar applies for when `k` is in the `reduced_deps`.

src/operations.jl
```diff
@@ -56,29 +79,83 @@ Base.:(==)(x::ArrayReferenceMeta, y) = false
...
+    "The loop variables this operation depends on"
     dependencies::Vector{Symbol}
+    "??"
     reduced_deps::Vector{Symbol}
+    "Operations whose result this operation depends on"
     parents::Vector{Operation}
+    "For `memload` or `memstore`, encodes the array location"
     ref::ArrayReferenceMeta
+    "??"
     mangledvariable::Symbol
```
chriselrod, line 155: Caches a `gensym(variable)` call.

chriselrod:
Thanks for the documentation improvements! I answered all the questions I saw.

This is my first library with documentation, so it's all still new to me.
I think it complains about `vfilter!` because:
```julia
if (Base.libllvm_version ≥ v"7" && VectorizationBase.AVX512F) || Base.libllvm_version ≥ v"9"
# vfilter! definition
end
```
[Definition of vfilter!](https://github.com/chriselrod/LoopVectorization.jl/blob/master/src/filter.jl).
`vfilter!` depends on the compressed store LLVM intrinsic that was added with LLVM 7.
This intrinsic maps to an AVX512 assembly instruction. There was no fallback definition, causing LLVM to crash when the host cpu didn't support it. As of LLVM 9, it works regardless of backend (just like `fma` is simulated to work as expected instead of crashing when the host CPU doesn't have `fma` instructions).

Crashes are unpleasant, hence the `if` surrounding the definition.

There should be a fallback that just calls `filter!`, as there is for `vfilter`.

A dependency on `DocStringExtensions` sounds good.

timholy:
Thanks for the great answers! It really helps a lot.

This could *almost* be merged, but I couldn't resist trying to figure out what a few more things did and you'll see I embedded additional questions. The questions are only in the second commit, and just concern the `@avx` macro itself (the meaning of its "keyword" options) and the meaning of the `T` parameter during loop reconstruction.

src/constructors.jl
```diff
@@ -124,7 +142,7 @@ function check_macro_kwarg(arg, inline::Int8 = Int8(2), U::Int8 = zero(Int8), T:
     @assert arg.head === :(=)
     i = check_inline(arg)
     if i !== nothing
-        inline = i ? Int8(2) : Int8(-1)
+        inline = i ? Int8(2) : Int8(-1)  # what do these constants mean and why not keep true/false?
```
chriselrod, line 145: There were once three possibilities, now there are two.

src/constructors.jl
```diff
@@ -90,6 +90,24 @@ julia> b ≈ c
...
+
+where `body` is the code of the block (e.g., `for ... end`).
+
+`inline` is a Boolean that ?doesn't do anything?.
+`unroll` is an integer that specifies the loop unrolling factor.
+`tile` a tuple of integers, one per loop (so `(2,2)` would be appropriate
+for a `for m ∈ 1:size(A,1), n ∈ 1:size(B,2)` loop). ?Do these specify the number
+of tiles used in the computation?
+
+`unroll` takes precedence over `tile` (why??).
```
chriselrod, line 110: It'd be a breaking change, but I think it'd be good to remove the `tile` keyword, and instead allow `unroll` to accept either `Int` or `NTuple{N,Int}` arguments, perhaps requiring `N=2` initially.

src/constructors.jl
```diff
@@ -90,6 +90,24 @@ julia> b ≈ c
...
+```
+@avx inline=false unroll=2 tile=(2,2) body
+```
+
+where `body` is the code of the block (e.g., `for ... end`).
+
+`inline` is a Boolean that ?doesn't do anything?.
+`unroll` is an integer that specifies the loop unrolling factor.
+`tile` a tuple of integers, one per loop (so `(2,2)` would be appropriate
+for a `for m ∈ 1:size(A,1), n ∈ 1:size(B,2)` loop). ?Do these specify the number
```
chriselrod, line 107: Currently, `tile` means to unroll 2 loops. So `(3,4)` would mean to unroll one loop by a factor of 3, and another by a factor of 4.

What do you think would be the best naming convention? Repeated in the comment following this one, I'd like to replace the `tile` keyword with letting `unroll` accept `tuple` arguments.

Then, `unroll` would refer to loop unrolling, regardless of the number of loops being unrolled (although for now, the only possibilities are 0, 1, and 2).
While a `tile` would be much larger, and refer to evaluating pieces of the underlying array at a time for better cache efficiency -- an optimization `LoopVectorization` does not do yet.

src/operations.jl
```diff
@@ -1,7 +1,18 @@
...
 
+$(TYPEDFIELDS)
+"""
 struct ArrayReference
+    "The array variable"
     array::Symbol
+    "The list of indices (e.g., `[:i, :j]`), or `name(op)` for computed indices."
     indices::Vector{Symbol}
+    "Index offset, e.g., `a[i+7]` would store the `7`. Only used for small (`Int8`) offsets."
     offsets::Vector{Int8}
```
chriselrod, line 16: To add, `offsets` is used currently to identify loops that may benefit from unrolling. For example, in the exponential smoothing example, where we have both `X[Ipre,i,Ipost]` and `X[Ipre,i-1,Ipost]`. Unrolling `i` lets us save on `X` reloads.

Later, we can use it to identify loop-carried dependencies.

It is possible someone writes a loop using `X[Ipre,i + 130,Ipost]` and `X[Ipre,i + 129,Ipost]`, but before we reply on it for loop-carried dependencies (where it's just an optimization instead of correctness issue), I think we can live without that. It would be more common for people writing loops to define the `i` loop to incorporate that `129` offset.
Maybe when finally adding loop carried dependency support, we should increase it to `Int16`, and use an `Int128` in `condense_loopsets` (to hold the 8 offsets).
Or switch to something more flexible.

chriselrod:
On master, I:
1. Renamed the `tile` keyword argument `unroll`.
2. Made `inline` a boolean.
3. Bumped the version to `0.7.0`, because `1.` is breaking.

> the meaning of the T parameter during loop reconstruction.

The `T` parameter is for unrolling a second loop. `U` and `T` are handled differently right now, but eventually, [I'll replace them with a vector of length equal to the number of loops](https://github.com/chriselrod/LoopVectorization.jl/issues/73). An entry of `1` would mean not unrolled.

It would probably be better to call them `U1` and `U2` for now, although they are still handled differently in the internal code: Lowering unrolls code `U` times, and gets called `max(1, T)` times.

src/constructors.jl
timholy: I see you already did that, many thanks! I don't mind merge commits but I could have also done that. (Easier for me that you did, thanks!)

src/constructors.jl
timholy: Great idea,and I see you already did it. Since it wasn't documented it's probably not essential to support the old syntax with a deprecation warning.

src/constructors.jl
timholy: I think you made the right decision.

I'm still a little confused about `inline`, in my tests it doesn't actually seem to change the generated code at all and https://github.com/chriselrod/LoopVectorization.jl/blob/6f299fabff64b756ff18538bf767cb85fed133e1/src/condense_loopset.jl#L402-L406 does not entirely resolve my confusion.

src/operations.jl
timholy: The "save reloads" is awesome. Such an impressive package!

I suspect `Int8` is big enough for the *vast* majority of stencils. I just wanted to document the restriction, but I don't think the restriction is likely to get in the way of practical applications.

src/constructors.jl
chriselrod: What were your tests?

Hmm -- maybe I should go back to three-possibilities:
1. `@inline`
2. `@noinline`
3. Let some cost model decide.

Currently, `inline=false` let's Julia's cost modeling decide. Perhaps I should make it `2.` instead.
Feel free to add a `pushfirst!(q.args, Expr(:meta,:noinline))` to `__avx__!` (note the double-underscores).

`_avx_!` (single underscores) is automatically inlined. With `inline = true` (the default), `@avx` generates a call to `_avx_!`.
With `inline=false`, it calls `__avx__!` instead. That's so that `__avx__!` can then take the pointers of arrays and calculate some offsets within the same function, hopefully providing LLVM with a little more information, but I'm sure if it really helps.
`_avx_!` itself currently expects stridedpointers. I thought that was the easiest/more trait-like way of supporting a variety of array types: they just have to define an appropriate `stridedpointer` method.

Because LoopVectorization generates a lot of code through all the loop unrolling and branches to handle remainders, I figured it is unlikely that `__avx__!` would be inlined.

For example:
```julia

julia> function mydot(a, b)
           s = zero(promote_type(eltype(a), eltype(b)))
           @inbounds @simd for i ∈ eachindex(a,b)
               s += a[i] * b[i]
           end
           s
       end
mydot (generic function with 1 method)

julia> function mydotavx(a, b)
           s = zero(promote_type(eltype(a), eltype(b)))
           @avx for i ∈ eachindex(a,b)
               s += a[i] * b[i]
           end
           s
       end
mydotavx (generic function with 1 method)

julia> function mydotavx_noinline(a, b)
           s = zero(promote_type(eltype(a), eltype(b)))
           @avx inline = false for i ∈ eachindex(a,b)
               s += a[i] * b[i]
           end
           s
       end
mydotavx_noinline (generic function with 1 method)

julia> a = rand(31); b = rand(31);

julia> @btime mydot($a, $b)
  15.786 ns (0 allocations: 0 bytes)
6.91914940245568

julia> @btime mydotavx($a, $b)
  6.513 ns (0 allocations: 0 bytes)
6.91914940245568

julia> @btime mydotavx_noinline($a, $b)
  6.838 ns (0 allocations: 0 bytes)
6.91914940245568

julia> M = K = N = 128
128

julia> A = rand(M, K); B = rand(K, N); C1 = Matrix{Float64}(undef, M, N); C2 = similar(C1);

julia> a = view(A, 1:31, 1);

julia> b = view(B, 1:31, 1);

julia> @btime mydot($a, $b)
  17.855 ns (0 allocations: 0 bytes)
7.1870499901207525

julia> @btime mydotavx($a, $b)
  6.540 ns (0 allocations: 0 bytes)
7.187049990120752

julia> @btime mydotavx_noinline($a, $b)
  12.832 ns (0 allocations: 0 bytes)
7.187049990120752

julia> a = view(A, 1:32, 1);

julia> b = view(B, 1:32, 1);

julia> @btime mydot($a, $b)
  4.884 ns (0 allocations: 0 bytes)
7.214710063797762

julia> @btime mydotavx($a, $b)
  6.076 ns (0 allocations: 0 bytes)
7.214710063797762

julia> @btime mydotavx_noinline($a, $b)
  13.052 ns (0 allocations: 0 bytes)
7.214710063797762

julia> a = view(A, 1:127, 1);

julia> b = view(B, 1:127, 1);

julia> @btime mydot($a, $b)
  28.757 ns (0 allocations: 0 bytes)
32.21951520775383

julia> @btime mydotavx($a, $b)
  12.314 ns (0 allocations: 0 bytes)
32.21951520775384

julia> @btime mydotavx_noinline($a, $b)
  18.822 ns (0 allocations: 0 bytes)
32.21951520775384

julia> a = view(A, :, 1);

julia> b = view(B, :, 1);

julia> @btime mydot($a, $b)
  8.329 ns (0 allocations: 0 bytes)
32.22429313653614

julia> @btime mydotavx($a, $b)
  11.668 ns (0 allocations: 0 bytes)
32.22429313653614

julia> @btime mydotavx_noinline($a, $b)
  14.647 ns (0 allocations: 0 bytes)
32.22429313653614

julia> a = rand(128); b = rand(128);

julia> @btime mydot($a, $b)
  8.350 ns (0 allocations: 0 bytes)
29.895720677897188

julia> @btime mydotavx($a, $b)
  8.760 ns (0 allocations: 0 bytes)
29.895720677897188

julia> @btime mydotavx_noinline($a, $b)
  9.596 ns (0 allocations: 0 bytes)
29.895720677897188
```
(Note that I'm being very generous to LoopVectorization by picking 31 as the length.)
I don't know why `inline = false` sometimes causes such a steep regression, nor why using `view`s do.

The margins in the [dot product benchmarks](https://chriselrod.github.io/LoopVectorization.jl/latest/examples/dot_product/) are very slim. That was the most important factor in deciding to make `inline=true` the default.

We can look at the native code to confirm that with `inline=false`, the loop was not inlined, instead calling `__avx__!`:
```asm
# julia> @code_native debuginfo=:none mydotavx_noinline(A, B)
        .text
        subq    $40, %rsp
        movq    %rdi, %r9
        movq    8(%rdi), %rax
        movq    %rax, %rcx
        sarq    $63, %rcx
        andnq   %rax, %rcx, %rax
        movq    8(%rsi), %rcx
        movq    %rcx, %rdx
        sarq    $63, %rdx
        andnq   %rcx, %rdx, %rcx
        cmpq    %rcx, %rax
        jne     L155
        movq    %rax, 8(%rsp)
        movq    %rsi, (%rsp)
        movabsq $"__avx__!", %rax
        movabsq $139712568057360, %rdi  # imm = 0x7F115DFD5E10
        movabsq $139712568057456, %rsi  # imm = 0x7F115DFD5E70
        movabsq $139712568058800, %rdx  # imm = 0x7F115DFD63B0
        movabsq $139712568058992, %rcx  # imm = 0x7F115DFD6470
        leaq    8(%rsp), %r8
        vxorpd  %xmm0, %xmm0, %xmm0
        callq   *%rax
        vextractf64x4   $1, %zmm0, %ymm1
        vaddpd  %zmm1, %zmm0, %zmm0
        vextractf128    $1, %ymm0, %xmm1
        vaddpd  %xmm1, %xmm0, %xmm0
        vpermilpd       $1, %xmm0, %xmm1 # xmm1 = xmm0[1,0]
        vaddsd  %xmm1, %xmm0, %xmm0
        addq    $40, %rsp
        vzeroupper
        retq
L155:
        movabsq $jl_system_image_data, %rax
        movq    %rax, 16(%rsp)
        movq    %r9, 24(%rsp)
        movq    %rsi, 32(%rsp)
        movabsq $jl_invoke, %rax
        movabsq $jl_system_image_data, %rdi
        leaq    16(%rsp), %rsi
        movabsq $139712388188304, %rcx  # imm = 0x7F115344C890
        movl    $3, %edx
        callq   *%rax
        ud2
```
while the default version is inlined:
```asm
# julia> @code_native debuginfo=:none mydotavx(A, B)
        .text
        subq    $24, %rsp
        movq    8(%rdi), %rax
        movq    %rax, %rcx
        sarq    $63, %rcx
        andnq   %rax, %rcx, %r8
        movq    8(%rsi), %rax
        movq    %rax, %rcx
        sarq    $63, %rcx
        andnq   %rax, %rcx, %rax
        cmpq    %rax, %r8
        jne     L622
        movq    (%rdi), %rdx
        movq    (%rsi), %rax
        leal    -1(%r8), %ecx
        andl    $7, %ecx
        movb    $-2, %sil
        shlb    %cl, %sil
        cmpq    $7, %rcx
        notb    %sil
        movzbl  %sil, %esi
        movl    $255, %r9d
        cmovnel %esi, %r9d
        leaq    -30(%r8), %rdi
        cmpq    $2, %rdi
        jl      L266
        vxorpd  %xmm0, %xmm0, %xmm0
        movl    $24, %ecx
        vxorpd  %xmm2, %xmm2, %xmm2
        vxorpd  %xmm3, %xmm3, %xmm3
        vxorpd  %xmm1, %xmm1, %xmm1
        nopl    (%rax,%rax)
L128:
        movq    %rcx, %rsi
        vmovupd -192(%rdx,%rcx,8), %zmm4
        vmovupd -128(%rdx,%rcx,8), %zmm5
        vmovupd -64(%rdx,%rcx,8), %zmm6
        vmovupd (%rdx,%rcx,8), %zmm7
        vfmadd231pd     -192(%rax,%rcx,8), %zmm4, %zmm1 # zmm1 = (zmm4 * mem) + zmm1
        vfmadd231pd     -128(%rax,%rcx,8), %zmm5, %zmm3 # zmm3 = (zmm5 * mem) + zmm3
        vfmadd231pd     -64(%rax,%rcx,8), %zmm6, %zmm2 # zmm2 = (zmm6 * mem) + zmm2
        vfmadd231pd     (%rax,%rcx,8), %zmm7, %zmm0 # zmm0 = (zmm7 * mem) + zmm0
        addq    $32, %rcx
        addq    $9, %rsi
        cmpq    %rdi, %rsi
        jl      L128
        addq    $-23, %rcx
        cmpq    %rcx, %r8
        jl      L563
L219:
        leaq    -8(%r8), %rsi
        cmpq    %rcx, %rsi
        jge     L298
        kmovd   %r9d, %k1
        vmovupd -8(%rdx,%rcx,8), %zmm4 {%k1} {z}
        vmovupd -8(%rax,%rcx,8), %zmm5 {%k1} {z}
        vfmadd231pd     %zmm5, %zmm4, %zmm1 {%k1} # zmm1 = (zmm4 * zmm5) + zmm1
        jmp     L563
L266:
        movl    $1, %ecx
        vxorpd  %xmm1, %xmm1, %xmm1
        vxorpd  %xmm3, %xmm3, %xmm3
        vxorpd  %xmm2, %xmm2, %xmm2
        vxorpd  %xmm0, %xmm0, %xmm0
        cmpq    %rcx, %r8
        jl      L563
        jmp     L219
L298:
        leaq    -16(%r8), %rsi
        cmpq    %rcx, %rsi
        jge     L367
        vmovupd -8(%rdx,%rcx,8), %zmm4
        kmovd   %r9d, %k1
        vmovupd 56(%rdx,%rcx,8), %zmm5 {%k1} {z}
        vmovupd 56(%rax,%rcx,8), %zmm6 {%k1} {z}
        vfmadd231pd     -8(%rax,%rcx,8), %zmm4, %zmm1 # zmm1 = (zmm4 * mem) + zmm1
        vfmadd231pd     %zmm6, %zmm5, %zmm3 {%k1} # zmm3 = (zmm5 * zmm6) + zmm3
        jmp     L563
L367:
        leaq    -24(%r8), %rsi
        cmpq    %rcx, %rsi
        jge     L455
        vmovupd -8(%rdx,%rcx,8), %zmm4
        vmovupd 56(%rdx,%rcx,8), %zmm5
        kmovd   %r9d, %k1
        vmovupd 120(%rdx,%rcx,8), %zmm6 {%k1} {z}
        vmovupd 120(%rax,%rcx,8), %zmm7 {%k1} {z}
        vfmadd231pd     -8(%rax,%rcx,8), %zmm4, %zmm1 # zmm1 = (zmm4 * mem) + zmm1
        vfmadd231pd     56(%rax,%rcx,8), %zmm5, %zmm3 # zmm3 = (zmm5 * mem) + zmm3
        vfmadd231pd     %zmm7, %zmm6, %zmm2 {%k1} # zmm2 = (zmm6 * zmm7) + zmm2
        jmp     L563
L455:
        addq    $-32, %r8
        cmpq    %rcx, %r8
        jge     L563
        vmovupd -8(%rdx,%rcx,8), %zmm4
        vmovupd 56(%rdx,%rcx,8), %zmm5
        vmovupd 120(%rdx,%rcx,8), %zmm6
        kmovd   %r9d, %k1
        vmovupd 184(%rdx,%rcx,8), %zmm7 {%k1} {z}
        vmovupd 184(%rax,%rcx,8), %zmm8 {%k1} {z}
        vfmadd231pd     -8(%rax,%rcx,8), %zmm4, %zmm1 # zmm1 = (zmm4 * mem) + zmm1
        vfmadd231pd     56(%rax,%rcx,8), %zmm5, %zmm3 # zmm3 = (zmm5 * mem) + zmm3
        vfmadd231pd     120(%rax,%rcx,8), %zmm6, %zmm2 # zmm2 = (zmm6 * mem) + zmm2
        vfmadd231pd     %zmm8, %zmm7, %zmm0 {%k1} # zmm0 = (zmm7 * zmm8) + zmm0
L563:
        vaddpd  %zmm0, %zmm3, %zmm0
        vaddpd  %zmm2, %zmm1, %zmm1
        vaddpd  %zmm0, %zmm1, %zmm0
        vextractf64x4   $1, %zmm0, %ymm1
        vaddpd  %zmm1, %zmm0, %zmm0
        vextractf128    $1, %ymm0, %xmm1
        vaddpd  %xmm1, %xmm0, %xmm0
        vpermilpd       $1, %xmm0, %xmm1 # xmm1 = xmm0[1,0]
        vaddsd  %xmm1, %xmm0, %xmm0
        addq    $24, %rsp
        vzeroupper
        retq
L622:
        movabsq $jl_system_image_data, %rax
        movq    %rax, (%rsp)
        movq    %rdi, 8(%rsp)
        movq    %rsi, 16(%rsp)
        movabsq $jl_invoke, %rax
        movabsq $jl_system_image_data, %rdi
        movq    %rsp, %rsi
        movabsq $139712388188304, %rcx  # imm = 0x7F115344C890
        movl    $3, %edx
        callq   *%rax
        ud2
```

For more expensive functions, the `inline=false` version can get better performance:
```julia

julia> M = K = N = 128
128

julia> function AmulBavx1!(C, A, B)
           @avx for m ∈ 1:size(A,1), n ∈ 1:size(B,2)
               Cₘₙ = zero(eltype(C))
               for k ∈ 1:size(A,2)
                   Cₘₙ += A[m,k] * B[k,n]
               end
               C[m,n] = Cₘₙ
           end
       end
AmulBavx1! (generic function with 1 method)

julia> function AmulBavx1_noinline!(C, A, B)
           @avx inline=false for m ∈ 1:size(A,1), n ∈ 1:size(B,2)
               Cₘₙ = zero(eltype(C))
               for k ∈ 1:size(A,2)
                   Cₘₙ += A[m,k] * B[k,n]
               end
               C[m,n] = Cₘₙ
           end
       end
AmulBavx1_noinline! (generic function with 1 method)

julia> A = rand(M, K); B = rand(K, N); C1 = Matrix{Float64}(undef, M, N); C2 = similar(C1);

julia> @benchmark AmulBavx1!($C2, $A, $B)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     37.724 μs (0.00% GC)
  median time:      37.882 μs (0.00% GC)
  mean time:        37.956 μs (0.00% GC)
  maximum time:     109.022 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark AmulBavx1_noinline!($C1, $A, $B)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     36.693 μs (0.00% GC)
  median time:      36.893 μs (0.00% GC)
  mean time:        36.951 μs (0.00% GC)
  maximum time:     120.853 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> C1 ≈ C2 ≈ A * B
true

julia> @code_native debuginfo=:none AmulBavx1_noinline!(C1, A, B) # not inlined
        .text
        pushq   %rbp
        movq    %rsp, %rbp
        pushq   %rbx
        andq    $-32, %rsp
        subq    $96, %rsp
        vxorps  %xmm0, %xmm0, %xmm0
        vmovaps %ymm0, (%rsp)
        movq    $0, 32(%rsp)
        movq    %rsi, 64(%rsp)
        movq    %fs:0, %rbx
        movq    $12, (%rsp)
        movq    -15712(%rbx), %rax
        movq    %rax, 8(%rsp)
        movq    %rsp, %rax
        movq    %rax, -15712(%rbx)
        movq    (%rsi), %r10
        movq    8(%rsi), %r9
        movq    16(%rsi), %rax
        movq    32(%rax), %rcx
        movq    24(%r9), %rdx
        movq    32(%r9), %rsi
        movq    %rdx, 40(%rsp)
        movq    %rcx, 48(%rsp)
        movq    %rsi, 56(%rsp)
        movq    %r9, 32(%rsp)
        movq    %rax, 24(%rsp)
        movq    %r10, 16(%rsp)
        movabsq $"__avx__!", %r11
        movabsq $139712396595344, %rdi  # imm = 0x7F1153C51090
        movabsq $139712396595440, %rsi  # imm = 0x7F1153C510F0
        movabsq $139712396597840, %rdx  # imm = 0x7F1153C51A50
        movabsq $139712396597936, %rcx  # imm = 0x7F1153C51AB0
        leaq    40(%rsp), %r8
        pushq   %r10
        pushq   %rax
        vzeroupper
        callq   *%r11
        addq    $16, %rsp
        movq    8(%rsp), %rax
        movq    %rax, -15712(%rbx)
        movabsq $jl_system_image_data, %rax
        leaq    -8(%rbp), %rsp
        popq    %rbx
        popq    %rbp
        retq
```

The cost of a function call is negligible on the order of microseconds.

You've done a lot of work with Julia's inliner, so you know better than I do what the inline cost modeling should look like. Any thoughts or suggestions?
I think using LoopVectorization's cost modeling to decide would be the best default (better than `inline = true`, anyway).

src/operations.jl
chriselrod: [This](https://github.com/chriselrod/LoopVectorization.jl/blob/07de1e49655bcae5cb1634416868bfe77d85240f/src/determinestrategy.jl#L461) is the part of `determinestrategy.jl` that looks for offsets.

Lowering also currently special cases when `tiled` is the loop that has the offset. LLVM will probably handle when `unrolled` is that loop, but I'll probably add special case handling to be sure.

5 sets of benchmarks (this should be on your CartesianIndexing compiler pass PR), setup
```julia
using LoopVectorization, BenchmarkTools, Test
function smoothdim!(s, x, α, Rpre, irng::AbstractUnitRange, Rpost)
    ifirst, ilast = first(irng), last(irng)
    ifirst > ilast && return s
    @inbounds @fastmath for Ipost in Rpost
        # Initialize the first value along the filtered dimension
        for Ipre in Rpre
            s[Ipre, ifirst, Ipost] = x[Ipre, ifirst, Ipost]
        end
        # Handle all other entries
        for i = ifirst+1:ilast
            for Ipre in Rpre
                s[Ipre, i, Ipost] = α*x[Ipre, i, Ipost] + (1-α)*x[Ipre, i-1, Ipost]
            end
        end
    end
    s
end
function smoothdim_avx!(s, x, α, Rpre, irng::AbstractUnitRange, Rpost)
    ifirst, ilast = first(irng), last(irng)
    ifirst > ilast && return s
    @avx for Ipost in Rpost
         for Ipre in Rpre
            s[Ipre, ifirst, Ipost] = x[Ipre, ifirst, Ipost]
            for i = ifirst+1:ilast
                s[Ipre, i, Ipost] = α*x[Ipre, i, Ipost] + (1-α)*x[Ipre, i-1, Ipost]
            end
        end
    end
    s
end
function smoothdim_ifelse_avx!(s, x, α, Rpre, irng::AbstractUnitRange, Rpost)
    ifirst, ilast = first(irng), last(irng)
    ifirst > ilast && return s
    @avx for Ipost in Rpost, i = ifirst:ilast, Ipre in Rpre
        xi = x[Ipre, i, Ipost]
        xim = i > ifirst ? x[Ipre, i-1, Ipost] : xi
        s[Ipre, i, Ipost] = α*xi + (1-α)*xim
    end
    s
end

M = 8;
x = rand(M,M,M,M,M);
dest1, dest2, dest3 = similar(x), similar(x), similar(x);
α = 0.3
```
Note that `M = 8`. With `M = 11`, this benchmark is mostly bandwidth limited, meaning `smoothdim_avx!` performs worst (except when the Base version shows excess allocations at `d == 1` and `d == 2`, as shown below), because `LoopVectorization` splits the loop into two separate passes.
For larger problems (like `M = 11`), given that it's bandwidth limited, cache-based optimizations would help more.

```julia
julia> d = 1
1

julia> Rpre  = CartesianIndices(axes(x)[1:d-1]);

julia> Rpost = CartesianIndices(axes(x)[d+1:end]);

julia> @benchmark smoothdim!($dest1, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  19.69 MiB
  allocs estimate:  598016
  --------------
  minimum time:     30.222 ms (0.00% GC)
  median time:      30.363 ms (0.00% GC)
  mean time:        31.355 ms (3.19% GC)
  maximum time:     34.358 ms (0.00% GC)
  --------------
  samples:          160
  evals/sample:     1

julia> @benchmark smoothdim_avx!($dest2, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     19.985 μs (0.00% GC)
  median time:      20.483 μs (0.00% GC)
  mean time:        20.514 μs (0.00% GC)
  maximum time:     44.879 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark smoothdim_ifelse_avx!($dest3, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     13.287 μs (0.00% GC)
  median time:      13.619 μs (0.00% GC)
  mean time:        13.645 μs (0.00% GC)
  maximum time:     52.820 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @test dest1 ≈ dest2 ≈ dest3
Test Passed
```
These excess allocations were the case on regular Julia master as well. Any idea where the allocations are coming from? `@code_warntype` looks fine, as far as I can tell.

When `M == 8`, LoopVectorization is at it's slowest when `d == 1`, because `d == 1` does not allow loads to be eliminated. That's because when `d == 1`, the `i` loop is vectorized
```julia
x1 = X[i - 1,...]
x2 = X[i    ,...]
x3 = X[i + W - 1,...]
x4 = X[i + W    ,...]
```
`x2 == x3` will only be true if `W = 1`, where `W` is the increment between subsequent loop iterations (and also the width of SIMD vectors).

When `M == 11`, this was actually the fastest loop, since LoopVectorization would traverse this loop nest in the most cache-friendly minor. Load-elimination doesn't help when memory-bandwidth-bottlenecked, because the memory of the load you're eliminating had to already be available for loading.

`d == 2` shows the same `Base` problem, but demonstrates the benefit of load elimination when the arrays are smaller:
```julia
julia> d = 2
2

julia> Rpre  = CartesianIndices(axes(x)[1:d-1]);

julia> Rpost = CartesianIndices(axes(x)[d+1:end]);

julia> @benchmark smoothdim!($dest1, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  19.69 MiB
  allocs estimate:  598016
  --------------
  minimum time:     30.178 ms (0.00% GC)
  median time:      30.279 ms (0.00% GC)
  mean time:        31.293 ms (3.15% GC)
  maximum time:     34.247 ms (0.00% GC)
  --------------
  samples:          160
  evals/sample:     1

julia> @benchmark smoothdim_avx!($dest2, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     8.142 μs (0.00% GC)
  median time:      8.339 μs (0.00% GC)
  mean time:        8.361 μs (0.00% GC)
  maximum time:     21.851 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     3

julia> @benchmark smoothdim_ifelse_avx!($dest3, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     9.478 μs (0.00% GC)
  median time:      9.849 μs (0.00% GC)
  mean time:        9.874 μs (0.00% GC)
  maximum time:     50.904 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @test dest1 ≈ dest2 ≈ dest3
Test Passed
```

The standard Julia version speeds up 1000-fold once the allocations are elliminated:
However, starting with `d == 3`:
```julia
julia> d = 3;

julia> Rpre  = CartesianIndices(axes(x)[1:d-1]);

julia> Rpost = CartesianIndices(axes(x)[d+1:end]);

julia> @benchmark smoothdim!($dest1, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     33.337 μs (0.00% GC)
  median time:      33.525 μs (0.00% GC)
  mean time:        33.587 μs (0.00% GC)
  maximum time:     59.428 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark smoothdim_avx!($dest2, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     7.335 μs (0.00% GC)
  median time:      7.569 μs (0.00% GC)
  mean time:        7.584 μs (0.00% GC)
  maximum time:     20.087 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     4

julia> @benchmark smoothdim_ifelse_avx!($dest3, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     10.104 μs (0.00% GC)
  median time:      10.498 μs (0.00% GC)
  mean time:        10.527 μs (0.00% GC)
  maximum time:     56.877 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @test dest1 ≈ dest2 ≈ dest3
Test Passed
```
With `d == 4`:
```julia
julia> d = 4;

julia> Rpre  = CartesianIndices(axes(x)[1:d-1]);

julia> Rpost = CartesianIndices(axes(x)[d+1:end]);

julia> @benchmark smoothdim!($dest1, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     32.990 μs (0.00% GC)
  median time:      33.333 μs (0.00% GC)
  mean time:        33.383 μs (0.00% GC)
  maximum time:     61.634 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark smoothdim_avx!($dest2, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     8.977 μs (0.00% GC)
  median time:      9.302 μs (0.00% GC)
  mean time:        9.332 μs (0.00% GC)
  maximum time:     78.718 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark smoothdim_ifelse_avx!($dest3, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     14.053 μs (0.00% GC)
  median time:      14.669 μs (0.00% GC)
  mean time:        14.700 μs (0.00% GC)
  maximum time:     52.863 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @test dest1 ≈ dest2 ≈ dest3
Test Passed
```
and finally, `d == 5`:
```julia
julia> d = 5;

julia> Rpre  = CartesianIndices(axes(x)[1:d-1]);

julia> Rpost = CartesianIndices(axes(x)[d+1:end]);

julia> @benchmark smoothdim!($dest1, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     33.267 μs (0.00% GC)
  median time:      33.487 μs (0.00% GC)
  mean time:        33.537 μs (0.00% GC)
  maximum time:     51.475 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark smoothdim_avx!($dest2, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     9.378 μs (0.00% GC)
  median time:      9.693 μs (0.00% GC)
  mean time:        9.727 μs (0.00% GC)
  maximum time:     81.404 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark smoothdim_ifelse_avx!($dest3, $x, $α, $Rpre, axes($x, $d), $Rpost)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     8.809 μs (0.00% GC)
  median time:      9.071 μs (0.00% GC)
  mean time:        9.087 μs (0.00% GC)
  maximum time:     22.969 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     3

julia> @test dest1 ≈ dest2 ≈ dest3
Test Passed
```

How large is `M` in practice?
If it tends to be much larger, iterating over statically sized tiles may help.

Static sizing would also help the Base Julia version, which wont benefit from SIMD otherwise unless the inner most loop is of length `size(x,1) >= 4*W`, where W is the SIMD vector width.

src/constructors.jl
timholy: Aha! I was expecting a difference in the code produced by `_avx_!`, so I stashed the outcome of its `avx_body` call in a `Ref` and then extracted it after making the call. I hadn't realized I was looking at it in the wrong way. The difference is quite clear with `@code_typed`.

Sorry, I should have figured that out, but I was so focused on trying to understand the two `setup` calls that I failed to notice that a key difference was in `generate_call`.

> You've done a lot of work with Julia's inliner, so you know better than I do what the inline cost modeling should look like. Any thoughts or suggestions?

The reality is that while the new inlining engine is vastly smarter than the old one (and generally makes better decisions), it's still often wrong. IMO the big problem is a failure to account for optimizations that can occur only when inlining; for example, if a block of code creates a view and then passes that view to a helper function, inlining the helper might allow Julia to elide the view-creation itself.  But that's not accounted for now. (It wouldn't be phenomenally difficult to implement, but I think there's every reason to expect that it would make our current compile latencies look like an eyeblink, and that seems like too high of a price to pay.)

I've pushed a new version and modified the PR title to indicate that I think this might be ready to merge.

src/reconstruct_loopset.jl
```diff
@@ -417,8 +417,26 @@ function _avx_loopset(OPSsv, ARFsv, AMsv, LPSYMsv, LBsv, vargs)
         AMsv, LPSYMsv, LBsv, vargs
     )
 end
+const _body_ = Ref{Any}(nothing)
```
chriselrod, line 420: This seems clever. Will make debugging easier.

chriselrod:
Please let me know if/when you think this is good to merge -- many thanks for the great documentation!

chriselrod:
I missed your earlier comment saying that you think it's good to merge, so I've gone ahead and merged.

src/reconstruct_loopset.jl
timholy: Oops, I didn't actually mean to commit that, but if you like it...

PR #95, opened 2020-04-13
A big renaming

timholy:
This does a partial rename of `U, T` to `u₁, u₂`. If the unicode is too annoying to type I can switch to `u1, u2`. I am sure this isn't complete, and sometimes it's not entirely consistent (some places I use `u₁loop` to mean the symbol, but in other cases `u₁loopsym`). But I thought I'd submit it and see if you like this or not.


src/constructors.jl
```diff
@@ -111,8 +111,12 @@ One can find some circumstances where `inline=true` is faster, and other circums
 where `inline=false` is faster, so the best setting may require experimentation.
 
 `unroll` is an integer that specifies the loop unrolling factor, or a
-tuple `(4, 2)` signaling that the generated code should unroll more than
-one loop.
+tuple `(u₁, u₂) = (4, 2)` signaling that the generated code should unroll more than
+one loop. `u₁` is the unrolling factor for the innermost loop and `u₂` for the next-innermost loop,
```
chriselrod, line 115: "inner most loop being unrolled" and "next inner most loop being unrolled". There may be loops that are inner with respect to these that are not getting unrolled.

src/determinestrategy.jl
```diff
@@ -239,110 +239,110 @@ function determine_unroll_factor(
...
 end
 
-function tile_cost(X, U, T, UL, TL)
-    Tfactor = (num_iterations(TL, T)/TL)
-    Ufactor = (num_iterations(UL, U)/UL)
-    # X[1]*Tfactor*Ufactor + X[4] + X[2] * Tfactor + X[3] * Ufactor
-    X[1] + X[4] + X[2] * Tfactor + X[3] * Ufactor
+function unroll_cost(X, u₁, u₂, u₁L, u₂L)
+    u₂factor = (num_iterations(u₂L, u₂)/u₂L)
+    u₁factor = (num_iterations(u₁L, u₁)/u₁L)
```
chriselrod, line 244: I just realized `num_iterations` should be replaced with `cld`.

timholy: Yes, it would compile faster, though I bet that's not a big item.

src/determinestrategy.jl
```diff
@@ -239,110 +239,110 @@ function determine_unroll_factor(
...
+    # X[1]*u₂factor*u₁factor + X[4] + X[2] * u₂factor + X[3] * u₁factor
+    X[1] + X[4] + X[2] * u₂factor + X[3] * u₁factor
 end
```
chriselrod, line 247: x[1] is the cost of operations that depend on u1loop and u2loop.
x[2] is the cost of operations that depend on u1loop.
x[3] is the cost of operations that depend on u2loop.
x[4] is the cost of operations that depend on neither.

By unrolling u2loop by a factor of u2, we get to reuse the results that only depend on u1loop u2 times, therefore we reduce the cost of the u1loop-dependent operations by roughly 1/u2.
Similar story for u1loop.
I don't recall why I didn't have `x[4]  * u₁factor * u₂factor`. I have to rethink that.
I think this is because operations didn't have a `reduced_children` when I first wrote this, so I just assumed the operations were being replicated in this part of the code.

Simply musing unrelated to this PR. I'll update it this and `solve_unroll` accordingly.

src/determinestrategy.jl
```diff
@@ -602,23 +602,23 @@ function evaluate_cost_tile(
         if isunrolled && istiled # no cost decrease; cost must be repeated
             cost_vec[1] += rt
             reg_pressure[1] += rp
-        elseif isunrolled # cost decreased by tiling
+        elseif isunrolled # cost decreased by u₁
```
chriselrod, line 605: depends on u1, cost decreased by unrolling u2loop.

timholy:
Thanks for the review. Just checking, is this something you'd actually appreciate or would it mostly be annoying? Be honest :smile:.

chriselrod:
Thanks, I do appreciate it!
I'd been saying I wanted to move towards u1/u2 instead of "unrolled" and "tiling", and then eventually to "u" being a vector. This is a step in the right direction.

Eventually, we can reclaim [tile](https://github.com/JuliaArrays/TiledIteration.jl) to mean larger chunks of an array.
Unrolling would be in the domain of microkernels (maximizing CPU efficiency, given that the memory is there), and tiling of the macrokernel (making sure the memory is there).

timholy:
Good to know. Do you like the unicode subscript or will that be too annoying to type during normal development?

chriselrod:
They look nice, and I tend to write unciode like `\in` and `\alpha`.

But I'm starting to wonder if I should prefer ascii in general (meaning replacing these as well), for the sake of search-ability.

timholy:
VSCode, at least, has no problem searching for the unicode.

chriselrod:
It doesn't sound like it'll be a problem. I tried `grep -nr` on the command line and `M-x rgrep` in emacs, both work fine (assuming you can copy the unicode you're searching for from somewhere).

chriselrod:
If you think this is good to go, I'll go ahead and merge this and then register another release sometime later today.

timholy:
Give me a bit yet and I'll fix this up a little more. Currently immersed in JuliaImages stuff...

timholy:
OK, I've fixed the issues you noted. I didn't go any farther with renaming, but at least this is progress.

PR #97, opened 2020-04-14
Add nospecialize and precompiles to reduce latency

timholy:
In my test this shaved about 6s off the time to run the tests (which might be within the noise, not sure, the total was around 350s). For a focused test like
```julia
using LoopVectorization, OffsetArrays

function avxgeneric!(out, A, kern, R=CartesianIndices(out), z=zero(eltype(out)))
    Rk = CartesianIndices(kern)
    @avx for I in R
        tmp = z
        for J in Rk
            tmp += A[I+J]*kern[J]
        end
        out[I] = tmp
    end
    out
end

A = rand(Float32, 100, 100);
kern = OffsetArray(rand(Float32, 3, 3), -1:1, -1:1);
out1 = OffsetArray(similar(A, size(A).-2), 1, 1);   # stay away from the edges of A
avxgeneric!(out1, A, kern)
```
it changed the total time (starting from the Linux command line) from about 10.8s to 10.0s. If you take away the time to load `LoopVectorization` itself, it's about 9.8s vs 8.8s.

So, it's a modest improvement. Feel free to take it or leave it, not a big deal either way for me.


chriselrod:
Looking at the precompile, I see `solve_tilesize` which is being renamed in https://github.com/chriselrod/LoopVectorization.jl/pull/95 .
What's the best approach to handling that / which PR do you want me to merge first?

timholy:
If you want this you can merge it first, I will fix in the other PR.

chriselrod:
Thanks, I appreciate a free 0.8-1s!

PR #155, opened 2020-10-22
Bump VectorizationBase compat

MasonProtter:
I need the ambiguity fixes from the latest VectorizationBase, running this to see if CI will pass.


MasonProtter:
Oh, duh. VectorizationBase 0.13 hasn't been released yet 🤦

chriselrod:
There are also a lot of breaking changes in VectorizationBase 0.13.
I figured I'd take the dive and make some more serious changes to LoopVectorization while updating it, which is one reason why it's taken a while to update.

PR #164, opened 2020-12-03
Print `versioninfo(; verbose = true)` at the beginning of the test suite

DilumAluthge:


chriselrod:
I think the failures are because of a problem on nightly, and not of this PR. I've checked it out locally and tests passed on master.

PR #167, opened 2020-12-06
Add Zygote gradient for `vmap`

mcabbott:
This is the other half of https://github.com/FluxML/Zygote.jl/pull/842, with the aim of not having Zygote depend directly on this package, just to define one gradient.

It's copied from the Zygote code, but simplified as there is no need to reverse. I can't figure out why that version is able to use `vmap` for the backwards pass, but this gives errors here. But before trying harder, does this sound like a good idea?

I don't think this can be done with ChainRulesCore yet (e.g. https://github.com/JuliaDiff/ChainRules.jl/issues/314), so it uses ZygoteRules. Besides that, it needs `unzip` and `accum` from Zygote, which I've copied here verbatim. Possibly not all their methods are needed. I've also just copied the tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,6 +12,7 @@ OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 SLEEFPirates = "476501e8-09a2-5ece-8869-fb82de89a1fa"
 UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 VectorizationBase = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
+ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
 ArrayInterface = "2.14.2"
@@ -27,6 +28,7 @@ julia = "1.5"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [targets]
-test = ["InteractiveUtils", "Random", "Test"]
+test = ["InteractiveUtils", "Random", "Test", "Zygote"]
--- a/src/LoopVectorization.jl
+++ b/src/LoopVectorization.jl
@@ -82,6 +82,7 @@ include("constructors.jl")
 include("user_api_conveniences.jl")
 include("mapreduce.jl")
 include("broadcast.jl")
+include("zygoterules.jl")
 
 """
 LoopVectorization provides macros and functions that combine SIMD vectorization and
--- a/src/zygoterules.jl
+++ b/src/zygoterules.jl
@@ -1,0 +1,64 @@
+
+import ZygoteRules # ChainRules doesn't yet allow calling AD inside a rule, like _pullback
+
+ZygoteRules.@adjoint function vmap(f, args::AbstractArray...)
+    ∇vmap(__context__, f, args...)
+end
+
+function ∇vmap(cx, f, args...)
+    ys_and_backs = vmap((xs...) -> ZygoteRules._pullback(cx, f, xs...), args...)
+    if isempty(ys_and_backs)
+        println("got empty case!")
+        return ys_and_backs, _ -> nothing
+    else
+        ys, backs = _zygote_unzip(ys_and_backs)
+        function back(Δ)
+            # Δf_and_args = _zygote_unzip(vmap((b, δ) -> b(δ), backs, Δ)) # MethodError: no method matching map!(::LoopVectorization.var"#601#606", ::Array{Tuple{Nothing,Float64},1}, ::Array{Zygote.var"#1848#back#217"{Zygote.var"#215#216"{Float64}},1}, ::Tuple{Float64,Float64,Float64})
+            Δf_and_args = _zygote_unzip(map((b, δ) -> b(δ), backs, Δ))
+            Δf = reduce(_zygote_accum, Δf_and_args[1])
+            (Δf, Δf_and_args[2:end]...)
+        end
+        back(::Nothing) = map(_ -> nothing, ys_and_backs)
+        return ys, back
+    end
+end
+
+# unzip -- from src/lib/array.jl
+
+struct _zygote_StaticGetter{i} end
+
+(::_zygote_StaticGetter{i})(v) where {i} = v[i]
+
+function _zygote_unzip(tuples)
+    N = length(first(tuples))
+    _zygote__unzip(tuples, Val(N))
+end
+
+@generated function _zygote__unzip(tuples, ::Val{N}) where {N}
+    Expr(:tuple, [:(map($(_zygote_StaticGetter{i}()), tuples)) for i in 1:N]...)
+end
+
+# accum -- from src/lib/lib.jl
+
+_zygote_accum() = nothing
+_zygote_accum(x) = x
+
+_zygote_accum(x, y) =
+  x === nothing ? y :
+  y === nothing ? x :
+  x + y
+
+_zygote_accum(x, y, zs...) = _zygote_accum(_zygote_accum(x, y), zs...)
+
+_zygote_accum(x::Tuple, y::Tuple) = _zygote_accum.(x, y)
+_zygote_accum(x::AbstractArray, y::AbstractArray) = _zygote_accum.(x, y)
+
+@generated function _zygote_accum(x::NamedTuple, y::NamedTuple)
+  grad(x) = x in fieldnames(y) ? :(y.$x) : :nothing
+  Expr(:tuple, [:($f=_zygote_accum(x.$f, $(grad(f)))) for f in fieldnames(x)]...)
+end
+
+function _zygote_accum(x::Base.RefValue, y::Base.RefValue)
+  @assert x === y
+  return x
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -39,6 +39,9 @@ const RUN_SLOW_TESTS = LoopVectorization.REGISTER_COUNT ≤ 16 || !parse(Bool, g
 
 @time @testset "LoopVectorization.jl" begin
     
+    @time include("zygoterules.jl")
+    # move down, later!
+
     @test isempty(detect_unbound_args(LoopVectorization))
 
     @time include("printmethods.jl")
--- a/test/zygoterules.jl
+++ b/test/zygoterules.jl
@@ -1,0 +1,45 @@
+using Zygote
+
+function ngradient(f, xs::AbstractArray...)
+  grads = zero.(xs)
+  for (x, Δ) in zip(xs, grads), i in 1:length(x)
+    δ = sqrt(eps())
+    tmp = x[i]
+    x[i] = tmp - δ/2
+    y1 = f(xs...)
+    x[i] = tmp + δ/2
+    y2 = f(xs...)
+    x[i] = tmp
+    Δ[i] = (y2-y1)/δ
+  end
+  return grads
+end
+
+function gradcheck(f, xs...)
+  grad_zygote = gradient(f, xs...)
+  grad_finite_difference = ngradient(f, xs...)
+  return all(isapprox.(grad_zygote, grad_finite_difference; rtol = 1e-5, atol = 1e-5))
+end
+
+gradtest(f, xs::AbstractArray...) = gradcheck((xs...) -> sum(sin.(f(xs...))), xs...)
+gradtest(f, dims...) = gradtest(f, rand.(Float64, dims)...)
+
+@testset "Zygote adjoint for $mapfunc" for mapfunc in [map, vmap] begin
+    @test gradtest(xs -> sum(mapfunc(x -> x^2, xs)), rand(2,3))
+    @test gradtest((xss...) -> sum(mapfunc((xs...) -> sqrt(sum(xs.^2)), xss...)), [rand(5) for _ in 1:6]...)
+    function foo(y)
+      bar = (x) -> x*y
+      sum(mapfunc(bar, 1:5))
+    end
+    @test gradtest(foo, 3)
+    # @test gradient(v -> sum([x for x in v]), [1.1,2.2,3.3]) == ([1, 1, 1],)
+  # end
+
+  # @testset "Tuple adjoint" begin
+    x = randn(3)
+    _, pb = Zygote.pullback(x -> mapfunc(abs2, x), x)
+    Δy = randn(3)
+    @test first(pb((Δy..., ))) ≈ first(pb(Δy))
+  # end
+end
+end


chriselrod:
Thanks for the PR!

As a quick aside, the `vmap` code doesn't depend on anything else from LoopVectorization, all it needs is VectorizationBase and SLEEFPirates.jl.
Once special functions are migrated from SLEEFPirates into VectorizationBase, I thought it'd make sense to move `vmap` and similar functions there as well.

But as of right now, if all someone wants to depend on is `vmap`, we could also split it off into it's own library to make it a lighter dependency.

I'll take a look at the code soon.

chriselrod:
A quick benchmark...
```julia
function ngradient(f::F, xs::Vararg{AbstractArray,K}) where {F,K}
    grads = zero.(xs)
    for (x, Δ) in zip(xs, grads), i in 1:length(x)
        δ = sqrt(eps())
        tmp = x[i]
        x[i] = tmp - δ/2
        y1 = f(xs...)
        x[i] = tmp + δ/2
        y2 = f(xs...)
        x[i] = tmp
        Δ[i] = (y2-y1)/δ
    end
    return grads
end

args = ntuple(_ -> rand(50) 6);

vfunc(xss::Vararg{Any,K}) where {K} = sum(vmap(@inline((xs...) -> log(sum(xs .^ 2))), xss...))
func(xss::Vararg{Any,K}) where {K} = sum(map(@inline((xs...) -> log(sum(xs .^ 2))), xss...))

@benchmark func($args...)
@benchmark vfunc($args...)

@benchmark ngradient(func, $args...)
@benchmark ngradient(vfunc, $args...)
@benchmark gradient(func, $args...)
@benchmark gradient(vfunc, $args...)
```
Yields:
```julia
julia> @benchmark func($args...)
BenchmarkTools.Trial:
  memory estimate:  496 bytes
  allocs estimate:  1
  --------------
  minimum time:     475.796 ns (0.00% GC)
  median time:      537.798 ns (0.00% GC)
  mean time:        627.108 ns (5.85% GC)
  maximum time:     22.073 μs (96.84% GC)
  --------------
  samples:          10000
  evals/sample:     196

julia> @benchmark vfunc($args...)
BenchmarkTools.Trial:
  memory estimate:  496 bytes
  allocs estimate:  1
  --------------
  minimum time:     187.928 ns (0.00% GC)
  median time:      211.574 ns (0.00% GC)
  mean time:        262.069 ns (13.18% GC)
  maximum time:     7.737 μs (95.96% GC)
  --------------
  samples:          10000
  evals/sample:     555

julia> @benchmark ngradient(func, $args...)
BenchmarkTools.Trial:
  memory estimate:  293.53 KiB
  allocs estimate:  606
  --------------
  minimum time:     280.414 μs (0.00% GC)
  median time:      285.803 μs (0.00% GC)
  mean time:        325.925 μs (6.26% GC)
  maximum time:     4.154 ms (81.53% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark ngradient(vfunc, $args...)
BenchmarkTools.Trial:
  memory estimate:  293.53 KiB
  allocs estimate:  606
  --------------
  minimum time:     119.619 μs (0.00% GC)
  median time:      123.160 μs (0.00% GC)
  mean time:        153.919 μs (12.89% GC)
  maximum time:     3.684 ms (94.66% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark gradient(func, $args...)
BenchmarkTools.Trial:
  memory estimate:  729.41 KiB
  allocs estimate:  16785
  --------------
  minimum time:     1.432 ms (0.00% GC)
  median time:      1.545 ms (0.00% GC)
  mean time:        1.682 ms (6.48% GC)
  maximum time:     9.039 ms (79.07% GC)
  --------------
  samples:          2968
  evals/sample:     1

julia> @benchmark gradient(vfunc, $args...)
BenchmarkTools.Trial:
  memory estimate:  726.58 KiB
  allocs estimate:  16838
  --------------
  minimum time:     1.395 ms (0.00% GC)
  median time:      1.521 ms (0.00% GC)
  mean time:        1.671 ms (6.61% GC)
  maximum time:     9.205 ms (75.78% GC)
  --------------
  samples:          2988
  evals/sample:     1
```
At this (relatively small? 300 parameter) size, finite differences has a 100x performance advantage.
I replaced `sqrt` with `log`, as `map` and `vmap` performed more or less the same in all cases with `sqrt`.

mcabbott:
I hadn't benchmarked anything, but it's a good question whether saving an array of backwards functions like this is the right approach at all here. For `map` over 10 big objects this is fine, but `vmap` is presumably useful for many tiny objects, where the cost is high.

(I guess I'm not entirely clear on how much is being saved for all these pullbacks, are they all the same type? They each close over some values, perhaps cheap if scalars. Maybe it's just that each one is slow?)

The one time I encountered something similar I ended up with SliceMap.jl, where `mapcols` saves every `pullback(f,x)` much like this, while `MapCols` instead uses ForwardDiff on each slice. This was a lot faster (although its lead has been reduces with more recent Julia versions):

https://github.com/mcabbott/SliceMap.jl/blob/master/docs/intro.md

That said, step 1 might be to something no worse here instead of Zygote, and step 2 doing better.

CarloLucibello:
> I hadn't benchmarked anything, but it's a good question whether saving an array of backwards functions like this is the right approach at all here. For map over 10 big objects this is fine, but vmap is presumably useful for many tiny objects, where the cost is high.
> (I guess I'm not entirely clear on how much is being saved for all these pullbacks, are they all the same type? They each close over some values, perhaps cheap if scalars. Maybe it's just that each one is slow?)

Is there any way to inspect the pullback for captured variables? If there are nonoe, we could just `vmap` the first pullback.
If there are captured variables we can use the current approach instead. Or find a way to deconstruct the pullback rule and turn it into something that can be mapped over the Deltas and the internal state

CarloLucibello:
actually accessing the internal state is quite simple:
```julia
julia> function f(x)
           s = x +2 
           return a -> s+a
       end
f (generic function with 1 method)

julia> g = f(3)
#11 (generic function with 1 method)

julia> g.s
5
```

mcabbott:
> Or find a way to deconstruct the pullback rule and turn it into something that can be mapped over the Deltas and the internal state

I haven't tried, but perhaps calling `gradient` repeatedly would be better than saving the results of `pullback`? Repeat the forward pass but don't pass these functions around.

Or using `Zygote.pushforward`? Sadly this does not even have a stub in ZygoteRules.jl, maybe we should fix that? I'm not sure how well it works in general, and haven't timed it here.

The issue is a lot like gradients for broadcasting, except simpler as `ndims` all agree. Zygote's generic fallback saves the results of `pullback`, there was talk of replacing it with forward mode.

chriselrod:
```julia
julia> using VectorizationBase, SLEEFPirates, ForwardDiff

julia> const W = VectorizationBase.pick_vector_width_val(Float64);

julia> vx = Vec(Core.VecElement.(ntuple(_ -> randn(), W)));

julia> d = ForwardDiff.Dual(vx, ForwardDiff.Partials((vbroadcast(W, 1.0),)))
Dual{Nothing}(Vec{8,Float64}<0.8119868415887304, 1.1556499730070482, 0.6440758795186547, -2.7059858414617466, 0.7600279952539798, -2.2057691971160813, -1.747253656352074, 1.6143737692943303>,Vec{8,Float64}<1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0>)

julia> sin(d)
Dual{Nothing}(Vec{8,Float64}<0.7256556680551465, 0.9150573023206725, 0.6004597174465653, -0.42196062348703683, 0.6889417368087974, -0.8050877888926317, -0.984471760374113, 0.9990506535004391>,Vec{8,Float64}<0.6880580291076032, 0.40332385680692534, 0.7996550054392151, -0.9066141584083225, 0.7248167239260404, -0.5931556728009717, -0.17554302329028934, -0.04356365159563629>)

julia> cos(d)
Dual{Nothing}(Vec{8,Float64}<0.6880580291076032, 0.40332385680692534, 0.7996550054392151, -0.9066141584083225, 0.7248167239260404, -0.5931556728009717, -0.17554302329028934, -0.04356365159563629>,Vec{8,Float64}<-0.7256556680551465, -0.9150573023206725, -0.6004597174465653, 0.42196062348703683, -0.6889417368087974, 0.8050877888926317, 0.984471760374113, -0.9990506535004391>)

julia> exp(d)
Dual{Nothing}(Vec{8,Float64}<2.252378663544724, 3.176087117674774, 1.9042264812255403, 0.06680443279318816, 2.1383360829206186, 0.11016575352565817, 0.17425184235314772, 5.024740291341217>,Vec{8,Float64}<2.252378663544724, 3.176087117674774, 1.9042264812255403, 0.06680443279318816, 2.1383360829206186, 0.11016575352565817, 0.17425184235314772, 5.024740291341217>)
```

May also make sense to work on ForwardDiff support?

mcabbott:
Since Zygote depends on ForwardDiff, it will always be loaded. If we move its utility functions for creating Dual numbers etc. into ZygoteRules, then I think they can be used here without the direct dependence: https://github.com/FluxML/ZygoteRules.jl/pull/14 . Would this be enough? 

To make `@tullio ys[i] := sin(xs[i]) grad=Dual` work (which is almost `vmap(sin, xs)` with ForwardDiff gradient) I think I had to overload a few other things: https://github.com/mcabbott/Tullio.jl/blob/master/src/grad/avxdual.jl#L44-L79 . Perhaps those will also be needed here, and should live somewhere else... then we are back to needing Requires.jl.

chriselrod:
Quick mock up:
```julia
using ForwardDiff, VectorizationBase, SLEEFPirates
using VectorizationBase: AbstractSIMD, AbstractStridedPointer, vadd


@generated function init_dual(v::Tuple{Vararg{AbstractSIMD,A}}) where {A}
    res = Expr(:tuple)
    q = Expr(:block, Expr(:meta,:inline))
    for a ∈ 1:A
        v_a = Symbol(:v_,a)
        push!(q.args, Expr(:(=), v_a, Expr(:ref, :v, a)))
        partials = Expr(:tuple)
        for i ∈ 1:A
            push!(partials.args, Expr(:call, i == a ? :one : :zero, v_a))
        end
        push!(res.args, :(ForwardDiff.Dual($v_a, ForwardDiff.Partials($partials))))
    end
    push!(q.args, res)
    q
end

@generated function dual_store!(∂p::Tuple{Vararg{AbstractStridedPointer,A}}, p::AbstractStridedPointer, ∂v, im::Vararg{Any,N}) where {A,N}
    quote
        $(Expr(:meta,:inline))
        v = ∂v.value
        ∂ = ∂v.partials
        VectorizationBase.vnoaliasstore!(p, v, im...)
        Base.Cartesian.@nexprs $A a -> VectorizationBase.vnoaliasstore!(∂p[a], ∂[a], im...)
        nothing
    end
end

function ∂vmap_singlethread!(
    f::F, ∂y::Tuple{Vararg{DenseArray{T},A}}, y::DenseArray{T},
    args::Vararg{<:DenseArray{<:Base.HWReal},A}
) where {F,T <: Base.HWReal, A}
    N = length(y)
    ptry = VectorizationBase.zero_offsets(stridedpointer(y))
    ptrargs = VectorizationBase.zero_offsets.(stridedpointer.(args))
    ptr∂y = VectorizationBase.zero_offsets.(stridedpointer.(∂y))
    
    i = 0
    V = VectorizationBase.pick_vector_width_val(T)
    W = Int(V)
    st = VectorizationBase.static_sizeof(T)
    zero_index = MM{W}(StaticInt(0), st)
    while i < N - ((W << 2) - 1)
        index = VectorizationBase.Unroll{1,1,4,1,W,0x0000000000000000}((i,))
        v = f(init_dual(vload.(ptrargs, index))...)
        dual_store!(ptr∂y, ptry, v, index)
        i = vadd(i, 4W)
    end
    while i < N - (W - 1) # stops at 16 when
        vᵣ = f(init_dual(vload.(ptrargs, ((MM{W}(i),),)))...)
        dual_store!(ptr∂y, ptry, vᵣ, (MM{W}(i),))
        i = vadd(i, W)
    end
    if i < N
        m = mask(T, N & (W - 1))
        dual_store!(ptry, f(init_dual(vload.(ptrargs, ((MM{W}(i),),), m))...), (MM{W}(i,),), m)
    end
    nothing
end
```
Example:
```julia
julia> foo(a, b) = log(a) + exp(b)
foo (generic function with 1 method)

julia> x = rand(200); y = randn(200);

julia> ∂x = similar(x); ∂y = similar(y); z = similar(x);

julia> @benchmark ∂vmap_singlethread!(foo, $partials, $z, $x, $y)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     399.861 ns (0.00% GC)
  median time:      401.361 ns (0.00% GC)
  mean time:        401.810 ns (0.00% GC)
  maximum time:     471.629 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     202

julia> ∂x ≈ ForwardDiff.gradient(x -> sum(foo.(x, y)), x)
true

julia> ∂y ≈ ForwardDiff.gradient(y -> sum(foo.(x, y)), y)
true
```

mcabbott:
> That should no longer be needed with VectorizationBase 0.13 (the version that switched from SVec to Vec).

Oh great, I will look into deleting it then.

I will also look more at what you did. My very naiive attempt using Dual via `seed` does run, but isn't fast yet. I guess `unzip` is pretty wasteful here, there's an intermediate vector of tuples, and perhaps `vmap` isn't happy about that. 

```julia
using Zygote, LoopVectorization
using Zygote: @adjoint, seed, extract, unzip, pullback

@adjoint vmap(f, args::AbstractArray...) = ∇vmap(f, args...)

function ∇vmap(f, xzs...)
    ys = vmap(f, xzs...)
    N = length(xzs)
    # duals = Tuple(seed(falses(N), Val(N)))
    duals = Tuple(seed(zeros(N), Val(N)))
    function back(Δs)
        backs = unzip(vmap(Δs, xzs...) do Δ, xz...
            Δ .* f((xz .+ duals)...).partials.values
        end)
        (nothing, backs...)
    end
    ys, back
end

rs = rand(3);
gradient(xs -> sum(map(sin, xs)), rs)
gradient(xs -> sum(vmap(sin, xs)), rs) # looks fine

foo(a, b) = log(a) + exp(b)
x = rand(200); y = randn(200); dz = randn(200);

_, b1 = pullback((x,y) -> map(foo, x, y), x, y)
_, b2 = pullback((x,y) -> vmap(foo, x, y), x, y)
_, b3 = pullback((x,y) -> broadcast(foo, x, y), x, y)

@btime b1($dz)  #  1.744 μs (8 allocations: 18.28 KiB)
@btime b2($dz)  #  3.625 μs (4 allocations: 6.81 KiB)
@btime b3($dz)  # 19.150 μs (1030 allocations: 32.61 KiB)
```

The other possibility is to use Zygote's own forward mode instead, but I haven't timed it yet.

chriselrod:
FWIW, benchmarks of both versions (much is the same as earlier, but I fixed a bug in `∂vmap_singlethread!`:
```julia
using ForwardDiff, VectorizationBase, SLEEFPirates
using VectorizationBase: AbstractSIMD, AbstractStridedPointer, vadd


@generated function init_dual(v::Tuple{Vararg{AbstractSIMD,A}}) where {A}
    res = Expr(:tuple)
    q = Expr(:block, Expr(:meta,:inline))
    for a ∈ 1:A
        v_a = Symbol(:v_,a)
        push!(q.args, Expr(:(=), v_a, Expr(:ref, :v, a)))
        partials = Expr(:tuple)
        for i ∈ 1:A
            push!(partials.args, Expr(:call, i == a ? :one : :zero, v_a))
        end
        push!(res.args, :(ForwardDiff.Dual($v_a, ForwardDiff.Partials($partials))))
    end
    push!(q.args, res)
    q
end

@generated function dual_store!(∂p::Tuple{Vararg{AbstractStridedPointer,A}}, p::AbstractStridedPointer, ∂v, im::Vararg{Any,N}) where {A,N}
    quote
        $(Expr(:meta,:inline))
        v = ∂v.value
        ∂ = ∂v.partials
        VectorizationBase.vnoaliasstore!(p, v, im...)
        Base.Cartesian.@nexprs $A a -> VectorizationBase.vnoaliasstore!(∂p[a], ∂[a], im...)
        nothing
    end
end
@generated function dual_store!(∂p::Tuple{Vararg{AbstractStridedPointer,A}}, p::AbstractStridedPointer, ∂v, im::Vararg{Any,N}) where {A,N}
    quote
        $(Expr(:meta,:inline))
        v = ∂v.value
        ∂ = ∂v.partials
        VectorizationBase.vnoaliasstore!(p, v, im...)
        Base.Cartesian.@nexprs $A a -> VectorizationBase.vnoaliasstore!(∂p[a], ∂[a], im...)
        nothing
    end
end

function ∂vmap_singlethread!(
    f::F, ∂y::Tuple{Vararg{DenseArray{T},A}}, y::DenseArray{T},
    args::Vararg{<:DenseArray{<:Base.HWReal},A}
) where {F,T <: Base.HWReal, A}
    N = length(y)
    ptry = VectorizationBase.zero_offsets(stridedpointer(y))
    ptrargs = VectorizationBase.zero_offsets.(stridedpointer.(args))
    ptr∂y = VectorizationBase.zero_offsets.(stridedpointer.(∂y))
    
    i = 0
    V = VectorizationBase.pick_vector_width_val(T)
    W = Int(V)
    st = VectorizationBase.static_sizeof(T)
    zero_index = MM{W}(StaticInt(0), st)
    while i < N - ((W << 2) - 1)
        index = VectorizationBase.Unroll{1,1,4,1,W,0x0000000000000000}((i,))
        v = f(init_dual(vload.(ptrargs, index))...)
        dual_store!(ptr∂y, ptry, v, index)
        i = vadd(i, 4W)
    end
    while i < N - (W - 1) 
        vᵣ = f(init_dual(vload.(ptrargs, ((MM{W}(i),),)))...)
        dual_store!(ptr∂y, ptry, vᵣ, (MM{W}(i),))
        i = vadd(i, W)
    end
    if i < N
        m = mask(T, N & (W - 1))
        dual_store!(ptr∂y, ptry, f(init_dual(vload.(ptrargs, ((MM{W}(i),),), m))...), (MM{W}(i,),), m)
    end
    nothing
end


using Zygote, LoopVectorization
using Zygote: @adjoint

@inline simd_map(f, args...) = vmap(f, args...)

struct SIMDMapBack{K,T<:Tuple{Vararg{Any,K}}}
    jacs::T
end
@generated function (b::SIMDMapBack{K,T})(Δ::A) where {K,T,A}
    t = Expr(:tuple, nothing)
    for k in 1:K
        push!(t.args, :(j[$k] .* Δ))
    end
    Expr(:block, :(j = b.jacs), t)
end

@adjoint function simd_map(f::F, _args::Vararg{Any,K}) where {F,K}
    args = first(_args)
    out = similar(first(args))
    jacs = map(similar, args)
    ∂vmap_singlethread!(f, jacs, out, args...)
    out, SIMDMapBack(jacs)
end


using Zygote: seed, unzip

@adjoint vmap(f, args::AbstractArray...) = ∇vmap(f, args...)

function ∇vmap(f, xzs...)
    ys = vmap(f, xzs...)
    N = length(xzs)
    # duals = Tuple(seed(falses(N), Val(N)))
    duals = Tuple(seed(zeros(N), Val(N)))
    function back(Δs)
        backs = unzip(vmap(Δs, xzs...) do Δ, xz...
            Δ .* f((xz .+ duals)...).partials.values
        end)
        (nothing, backs...)
    end
    ys, back
end

rs = rand(3);
gradient(xs -> sum(map(sin, xs)), rs)
gradient(xs -> sum(vmap(sin, xs)), rs) # looks fine
gradient(xs -> sum(simd_map(sin, xs)), rs) # looks fine

foo(a, b) = log(a) + exp(b)
x = rand(200); y = randn(200); dz = randn(200);

_, b1 = pullback((x,y) -> map(foo, x, y), x, y);
_, b2 = pullback((x,y) -> vmap(foo, x, y), x, y);
_, b3 = pullback((x,y) -> broadcast(foo, x, y), x, y);
_, b4 = pullback((x,y) -> simd_map(foo, x, y), x, y);

@btime b1($dz);  #  1.744 μs (8 allocations: 18.28 KiB)
@btime b2($dz);  #  3.625 μs (4 allocations: 6.81 KiB)
@btime b3($dz);  # 19.150 μs (1030 allocations: 32.61 KiB)
@btime b4($dz); 

g1 = @btime gradient(xy -> sum(map(foo, xy[1], xy[2])), (x, y));
g2 = @btime gradient(xy -> sum(vmap(foo, xy[1], xy[2])), (x, y));
g3 = @btime gradient(xy -> sum(simd_map(foo, xy[1], xy[2])), (x, y));

all(map(≈, only(g1), only(g2)))
all(map(≈, only(g1), only(g3)))
```

I get:
```julia
julia> rs = rand(3);

julia> gradient(xs -> sum(map(sin, xs)), rs)
([0.7200866240253904, 0.9646572575851471, 0.6962408149200021],)

julia> gradient(xs -> sum(vmap(sin, xs)), rs) # looks fine
([0.7200866240253904, 0.9646572575851471, 0.6962408149200021],)

julia> gradient(xs -> sum(simd_map(sin, xs)), rs) # looks fine
([0.7200866240253904, 0.9646572575851472, 0.6962408149200022],)

julia> foo(a, b) = log(a) + exp(b)
foo (generic function with 1 method)

julia> x = rand(200); y = randn(200); dz = randn(200);

julia> _, b1 = pullback((x,y) -> map(foo, x, y), x, y);

julia> _, b2 = pullback((x,y) -> vmap(foo, x, y), x, y);

julia> _, b3 = pullback((x,y) -> broadcast(foo, x, y), x, y);

julia> _, b4 = pullback((x,y) -> simd_map(foo, x, y), x, y);

julia> @btime b1($dz);  #  1.744 μs (8 allocations: 18.28 KiB)
  1.779 μs (8 allocations: 18.28 KiB)

julia> @btime b2($dz);  #  3.625 μs (4 allocations: 6.81 KiB)
  2.032 μs (4 allocations: 6.81 KiB)

julia> @btime b3($dz);  # 19.150 μs (1030 allocations: 32.61 KiB)
  13.866 μs (1027 allocations: 32.52 KiB)

julia> @btime b4($dz);
  286.367 ns (3 allocations: 3.56 KiB)

julia> g1 = @btime gradient(xy -> sum(map(foo, xy[1], xy[2])), (x, y));
  7.176 μs (47 allocations: 35.91 KiB)

julia> g2 = @btime gradient(xy -> sum(vmap(foo, xy[1], xy[2])), (x, y));
  4.601 μs (51 allocations: 11.38 KiB)

julia> g3 = @btime gradient(xy -> sum(simd_map(foo, xy[1], xy[2])), (x, y));
  1.602 μs (32 allocations: 10.17 KiB)

julia> all(map(≈, only(g1), only(g2)))
true

julia> all(map(≈, only(g1), only(g3)))
true
```
The pullback using `∂vmap_singlethread!` itself is very fast, but performance of the entire gradient is still bad/mostly overhead.
The sum and gradient with respect to it should be comparatively very fast, not take 4x longer.
Still, even just the other pullbacks by themselves were much slower.

chriselrod:
What do you think we should do here?

The `simd_map` implementation is several times faster than the alternatives, so I am in favor of it.

I'd be okay with using `Requires.jl`, and then placing a gradient definition behind it. What would be the best thing to look for? Perhaps `Zygote` itself, and then use `Zygote.ForwardDiff` within the code?

Feel free to do that here, or one of us can open a separate PR.

chriselrod:
@mcabbott Do you want to finish this, or I should I open a new PR with my implementation?

mcabbott:
I haven't circled back to this, I'm sorry. If you have a plan for what you'd like then don't hold back for me.

chriselrod:
Modified from [the test suite](https://github.com/chriselrod/LoopVectorization.jl/blob/master/test/zygote.jl):
```julia
julia> using LoopVectorization, Zygote, Test

julia> T = Float32
Float32

julia> x = randn(T, 217); y = randn(T, 217); z = randn(T, 217);

julia> # Test 1-arg and `tanh_fast` vs `tanh`
       gtref = gradient(x -> sum(map(tanh, x)), x);

julia> gtlv = gradient(x -> sum(vmap(tanh_fast, x)), x);

julia> @test  only(gtref) ≈ only(gtlv)
Test Passed

julia> # Test multiple arguments
       gref = gradient(xyz -> sum(map(zygotefun, xyz...)), (x, y, z));

julia> glv = gradient(xyz -> sum(vmap(zygotefun, xyz...)), (x, y, z));

julia> @test all(map(≈, only(gref), only(glv)))
Test Passed
```
Benchmarks:
```julia
julia> @benchmark gradient(x -> sum(map(tanh, x)), $x)
BenchmarkTools.Trial:
  memory estimate:  9.56 KiB
  allocs estimate:  34
  --------------
  minimum time:     2.632 μs (0.00% GC)
  median time:      2.803 μs (0.00% GC)
  mean time:        3.543 μs (12.09% GC)
  maximum time:     295.795 μs (96.88% GC)
  --------------
  samples:          10000
  evals/sample:     9

julia> @benchmark gradient(x -> sum(vmap(tanh_fast, x)), $x)
BenchmarkTools.Trial:
  memory estimate:  1.98 KiB
  allocs estimate:  3
  --------------
  minimum time:     206.505 ns (0.00% GC)
  median time:      282.190 ns (0.00% GC)
  mean time:        361.556 ns (17.17% GC)
  maximum time:     4.138 μs (85.45% GC)
  --------------
  samples:          10000
  evals/sample:     590

julia> @benchmark gradient(xyz -> sum(map(zygotefun, xyz...)), ($x, $y, $z))
BenchmarkTools.Trial:
  memory estimate:  38.59 KiB
  allocs estimate:  37
  --------------
  minimum time:     9.000 μs (0.00% GC)
  median time:      10.509 μs (0.00% GC)
  mean time:        13.253 μs (14.55% GC)
  maximum time:     3.044 ms (95.16% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark gradient(xyz -> sum(vmap(zygotefun, xyz...)), ($x, $y, $z))
BenchmarkTools.Trial:
  memory estimate:  3.95 KiB
  allocs estimate:  5
  --------------
  minimum time:     539.247 ns (0.00% GC)
  median time:      640.868 ns (0.00% GC)
  mean time:        823.896 ns (16.00% GC)
  maximum time:     13.667 μs (87.97% GC)
  --------------
  samples:          10000
  evals/sample:     182
```

PR #171, opened 2020-12-12
Add a test that will fail if the test suite takes more than two hours

### LINKED ISSUE #160: Test suite takes three hours to run on Julia nightly/master, but only 47 minutes to run on Julia 1.5.3
On Julia 1.5.3, the test suite takes 47 minutes to run: https://github.com/chriselrod/LoopVectorization.jl/runs/1475169996

However, on Julia nightly/master, the test suite takes three hours to run: https://github.com/chriselrod/LoopVectorization.jl/runs/1475170020

GitHub Actions allows each job in a workflow to run for [up to 6 hours of execution time](https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration). So we are under the limit there.

However, it is still concerning to see the duration of the test suite increase so drastically from Julia 1.5 to Julia nightly.

DilumAluthge:
Fixes #160 

Right now, if the test suite starts to take a really long time to run, we have no way of noticing (unless we happen to open up the test log).

This pull request adds a single test to the end of the test suite. This test will:
- PASS if the test suite takes two hours or less 
- FAIL if the test suite takes more than two hours

We can certainly change this limit. I picked two hours because, as far as I can tell, none of the jobs take more than 1.5 hours to run.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,13 +1,11 @@
 using Test
 using LoopVectorization
 using LinearAlgebra
-
 import InteractiveUtils
 
 InteractiveUtils.versioninfo(stdout; verbose = true)
 
-# const START_TIME = time()
-# exceeds_time_limit() = (time() - START_TIME) > 35 * 60
+const START_TIME = time()
 
 function clenshaw(x, coeff)
     len_c = length(coeff)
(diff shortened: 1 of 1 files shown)


chriselrod:
I think that's fine. I'll merge once tests pass.

DilumAluthge:
SGTM.

Are there any other packages for which we might want to add a similar "test suite total time" test?

chriselrod:
Maybe PaddedMatrices? The tests are very slow when getting coverage (but reasonably fast when not getting coverage, e.g. when running locally).

DilumAluthge:
> Maybe PaddedMatrices? The tests are very slow when getting coverage (but reasonably fast when not getting coverage, e.g. when running locally).

Yeah the coverage overhead can be brutal sometimes.

https://github.com/chriselrod/PaddedMatrices.jl/pull/9

PR #172, opened 2020-12-13
Remove `println` statements accompanying errors

mcabbott:
This replaces `println(ex); throw(str)` with an error type, so that if macro expansion fails within a try-catch block, nothing is printed. 

Closes https://github.com/mcabbott/Tullio.jl/issues/26

Example of new behaviour:
```
julia> try                                                                   
       macroexpand(Main, quote                                               
       @avx for i = 𝒶𝓍i                                                      
           if i ∈ (axes)(var"𝛥≪rand(10)≫", 1)                                
               var"𝛥≪rand(10)≫"[i] = var"𝛥≪rand(10)≫"[i] + 𝛥ℛ[i]             
           else                                                              
               (zero)((eltype)(var"𝛥≪rand(10)≫"))                            
           end                                                               
       end                                                                   
       end)                                                                  
       catch err                                                             
       end  # silence!

julia> @macroexpand @avx for i = 𝒶𝓍i                                                      
           if i ∈ (axes)(var"𝛥≪rand(10)≫", 1)                                
               var"𝛥≪rand(10)≫"[i] = var"𝛥≪rand(10)≫"[i] + 𝛥ℛ[i]             
           else                                                              
               (zero)((eltype)(var"𝛥≪rand(10)≫"))                            
           end                                                               
       end  # similar output to before:
ERROR: LoadError: Don't know how to handle expression.
if i ∈ axes(var"𝛥≪rand(10)≫", 1)
    #= REPL[10]:3 =#
    var"𝛥≪rand(10)≫"[i] = var"𝛥≪rand(10)≫"[i] + 𝛥ℛ[i]
else
    #= REPL[10]:5 =#
    zero(eltype(var"𝛥≪rand(10)≫"))
end
Stacktrace:
 [1] push!(ls::LoopVectorization.LoopSet, ex::Expr, elementbytes::Int64, position::Int64)
   @ LoopVectorization ~/.julia/dev/LoopVectorization/src/graphs.jl:754
 [2] add_block!
...
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LoopVectorization"
 uuid = "bdcacae8-1622-11e9-2a5c-532679323890"
 authors = ["Chris Elrod <elrodc@gmail.com>"]
-version = "0.9.6"
+version = "0.9.7"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
--- a/src/graphs.jl
+++ b/src/graphs.jl
@@ -526,7 +526,7 @@ function register_single_loop!(ls::LoopSet, looprange::Expr)
         U = add_loop_bound!(ls, itersym, Expr(:call, lv(:maybestaticlast), N), true)
         loop = Loop(itersym, L, U)
     else
-        throw("Unrecognized loop range type: $r.")
+        throw(LoopError("Unrecognized loop range type: $r."))
     end
     add_loop!(ls, loop, itersym)
     nothing
@@ -819,6 +819,6 @@ struct LoopError <: Exception
 end
 
 function Base.showerror(io::IO, err::LoopError)
-    printstyled(io, err.msg, '\n'; color = :red)
-    printstyled(io, err.ex)
+    printstyled(io, err.msg; color = :red)
+    isnothing(err.ex) || printstyled(io, '\n', err.ex)
 end
--- a/src/graphs.jl
+++ b/src/graphs.jl
@@ -621,8 +621,7 @@ function add_operation!(
     elseif RHS.head === :block
         add_operation!(ls, LHS, strip_op_linenumber_nodes(RHS), elementbytes, position)
     else
-        println(RHS)
-        throw("Expression not recognized.")
+        throw(LoopError("Expression not recognized.", RHS))
     end
 end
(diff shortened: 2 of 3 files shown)


chriselrod (approved):
Looks good to me.
Feel free to bump the patch version.

mcabbott:
Great, done!

chriselrod:
I'll merge and issue another release once tests pass.

chriselrod:
Perhaps add a
```julia
@test_throws LoadError @macroexpand @avx for i = 𝒶𝓍i                                                      
    if i ∈ (axes)(var"𝛥≪rand(10)≫", 1)                                
        var"𝛥≪rand(10)≫"[i] = var"𝛥≪rand(10)≫"[i] + 𝛥ℛ[i]             
    else                                                              
        (zero)((eltype)(var"𝛥≪rand(10)≫"))                            
    end                                                               
end
```
?

mcabbott:
Done. Assuming that this kind of if-else is something unlikely to ever work, I guess. It was just the first one I happened to pick up for quick tests locally.

By the way, there isn't an AVX-friendly way of doing that, is there? I mean of testing some index & returning zero (or some other padding) out of bounds? Something like `ifelse(i in axes(ys,1), ys[i], 0.0)` will actually read out of bounds

chriselrod:
Yeah, maybe it's better to have a wonkier example now, but I figured it wouldn't be that bad to change it later.

> By the way, there isn't an AVX-friendly way of doing that, is there? I mean of testing some index & returning zero (or some other padding) out of bounds? Something like `ifelse(i in axes(ys,1), ys[i], 0.0)` will actually read out of bounds

There is, see [here](https://github.com/chriselrod/LoopVectorization.jl/blob/b15676932d956a4f27ce8596b1b4e594496990a0/test/miscellaneous.jl#L1067) for an example.
If you use `?` instead of `ifelse`, LoopVectorization will convert that into a masked load.
So you'd want to do something similar to the linked line.
Instead of
```julia
xim = i > ifirst ? x[Ipre, i-1, Ipost] : xi
```
you'd want
```julia
((i >= first(axes(ys,1))) & (i <= last(axes(ys,1)))) ? ys[i] : 0.0
```
and that should work.

AVX has a masked move instruction:
https://www.felixcloutier.com/x86/vmaskmov
While with AVX512, it of course uses a regular move but applies a mask.

mcabbott:
That looks great, thanks! Will try this out.

PR #173, opened 2020-12-14
Increase the `ELAPSED_MINUTES` test limit from two hours to three hours

DilumAluthge:
On the most recent commit to `master`, one job took [2h 20m 9s](https://github.com/chriselrod/LoopVectorization.jl/runs/1545836570) and another job took [2h 12m 46s](https://github.com/chriselrod/LoopVectorization.jl/runs/1546008232).

So let's increase the limit from two hours to three hours.

(In the commit message, I wrote "two hours to four hours", which is a typo. As seen in the diff, the actual change in "two hours to three hours".)

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -79,4 +79,4 @@ const RUN_SLOW_TESTS = LoopVectorization.REGISTER_COUNT ≤ 16 || !parse(Bool, g
 end
 
 const ELAPSED_MINUTES = (time() - START_TIME)/60
-@test ELAPSED_MINUTES < 120
+@test ELAPSED_MINUTES < 180


chriselrod (approved):
As I said elsewhere, this slow down was because of a regression where some of the slower tests started using a fallback loop instead of LoopVectorization. This was fixed in ArrayInterface, causing a slowdown (because of longer compilation) when using the latest version of ArrayInterface.

So while not great to increase the limit to 3 hours, it at least isn't a sign of a recent regression in the ecosystem.

DilumAluthge:
> a regression where some of the slower tests started using a fallback loop instead of LoopVectorization

I wonder: would it be good to have an option in which if LoopVectorization could not be used, an exception would be thrown, instead of using a fallback loop? This option might make it easier to detect regressions like the ArrayInterface regression you fixed recently.

chriselrod:
That would be helpful for regression testing.

And as always, thanks for the PR.

PR #177, opened 2020-12-22
Replace Unicode subscript m (resp. n) with regular m (resp. n)

DilumAluthge:
Since some users (e.g. macOS users with only the default system fonts) might not be able to render the subscripts.


DilumAluthge:
This did not come out very well. Maybe we should only do it in the docs.

DilumAluthge:
Closing in favor of #179

PR #179, opened 2020-12-22
Docs only: Replace Unicode subscript i (resp. j, l, m, n, s) with _i (resp. _j, _l, _m, _n, _s)

DilumAluthge:
Since some users (e.g. macOS users with only the default system fonts) might not be able to render the subscripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,11 +239,11 @@ Note that loops will be faster than broadcasting in general. This is because the
 ```julia
 julia> function AmulBtest!(C,A,Bk,Bn,d)
           @avx for m ∈ axes(A,1), n ∈ axes(Bk,2)
-             ΔCₘₙ = zero(eltype(C))
+             ΔC_m_n = zero(eltype(C))
              for k ∈ axes(A,2)
-                ΔCₘₙ += A[m,k] * (Bk[k,n] + Bn[n,k])
+                ΔC_m_n += A[m,k] * (Bk[k,n] + Bn[n,k])
              end
-             C[m,n] = ΔCₘₙ + d[m]
+             C[m,n] = ΔC_m_n + d[m]
           end
        end
 AmulBtest! (generic function with 1 method)
--- a/docs/src/examples/array_interface.md
+++ b/docs/src/examples/array_interface.md
@@ -15,11 +15,11 @@ using StaticArrays, LoopVectorization
 
 @inline function AmulB!(C, A, B)
     @avx for n ∈ axes(C,2), m ∈ axes(C,1)
-        Cₘₙ = zero(eltype(C))
+        C_m_n = zero(eltype(C))
         for k ∈ axes(B,1)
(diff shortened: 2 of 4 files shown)


DilumAluthge:
I'll be honest, I'm not sure if this is actually an improvement. As @MasonProtter pointed out in a similar PR in Gaius, it makes the code harder to read.

DilumAluthge:
The original motivation for this is that I went to https://chriselrod.github.io/LoopVectorization.jl/latest/examples/matrix_multiplication/#Matrix-Multiplication on macOS, and I couldn't render the subscript m and subscript n in any of my three browsers (Firefox, Chrome, Safari). So the kernel was kind of hard to read because I couldn't tell what the subscripts were.

I've now installed a bunch more fonts on my computer, and now the subscripts render fine in both Firefox and Chrome.

The concern is this: if a user with only the default macOS system fonts goes to the above URL, will they still have troubles?

And more importantly, is it worth making the code harder to read to account for this edge case?

DilumAluthge:
It honestly might be better to just instruct users on how to install an appropriate font.

chriselrod:
I'm normally err on the side of accessibility over visual pizazz., but okay.
Is there a good resource somewhere instructing people on how to install fonts? If so, it'd be nice to then provide a link somewhere.

I was just waiting for the tests to finish before merging.

DilumAluthge:
> I'm normally err on the side of accessibility over visual pizazz., but okay.
> Is there a good resource somewhere instructing people on how to install fonts? If so, it'd be nice to then provide a link somewhere.
> 
> I was just waiting for the tests to finish before merging.

It's entirely up to you!

DilumAluthge:
>  Is there a good resource somewhere instructing people on how to install fonts? If so, it'd be nice to then provide a link somewhere.

Oh boy.... for me, it took a whole Slack thread with three people guiding me through it 😂 

I wouldn't describe installing fonts on macOS as a particularly intuitive process.

MasonProtter:
I feel like writing `Cij` is better than `C_i_j` if we're going to remove the unicode.

chriselrod:
I also think `Cij` looks a little better than `C_i_j`, but my preference isn't that strong so I don't really mind merging this anyway.

Re getting rid of the unicode:
I also texted a link to the docs to a friend with a Mac.
He said it looked fine except that the spacing was off and send a screen shot, clearly showing the unicode wasn't rendering. I do think it's better to get rid of it.

DilumAluthge:
> I feel like writing `Cij` is better than `C_i_j` if we're going to remove the unicode.

Sounds good. 

It's easiest for me to first revert this PR (https://github.com/chriselrod/LoopVectorization.jl/pull/186), and then do the replacements to change the unicode to `Cij`, etc.

DilumAluthge:
https://github.com/chriselrod/LoopVectorization.jl/pull/188

PR #180, opened 2020-12-24
Some inferrability and precompile improvements

timholy:
This is the companion to https://github.com/chriselrod/VectorizationBase.jl/pull/29. This PR focuses on inference improvements, as I've found for some packages that this is one of the most effective tools for increasing its precompilability (xref https://github.com/JuliaLang/julia/issues/30488#issuecomment-750783623). That said, you were already in quite good shape here, so these are minor tweaks with relatively little impact. The new precompile statements make the most difference, and those are something that the old `@snoopi` couldn't have easily discovered since they represent an intermediate stage in the inference tree. (In retrospect they seem pretty obvious, but when profiling it never ceases to amaze me how often things only make sense in retrospect and are completely different from what I expected going in.)


src/add_compute.jl
```diff
@@ -303,7 +303,7 @@ end
 
 # adds x ^ (p::Real)
 function add_pow!(
-    ls::LoopSet, var::Symbol, x, p::Real, elementbytes::Int, position::Int
+    ls::LoopSet, var::Symbol, @nospecialize(x), p::Real, elementbytes::Int, position::Int
```
DilumAluthge, line 306: Will adding this `@nospecialize` have an effect on runtime performance?

DilumAluthge, line 306: Will adding this `@nospecialize` have an effect on runtime performance?

src/constructors.jl
```diff
@@ -143,6 +143,7 @@ use their `parent`. Triangular loops aren't yet supported.
 """
 macro avx(q)
     q = macroexpand(__module__, q)
+    isa(q, Expr) || return q
```
DilumAluthge, line 146: Can you explain this line?

src/add_compute.jl
DilumAluthge, line 306: Will adding this `@nospecialize` have an effect on runtime performance?

src/vectorizationbase_compat/contract_pass.jl
```diff
@@ -107,14 +107,14 @@ function recursive_muladd_search!(call, argv, cnmul::Bool = false, csub::Bool =
...
+                end
             end
         end
     end
     length(call.args) == 4, cnmul, csub
 end
-                          
-function capture_muladd(ex::Expr, mod, LHS = nothing)
+
+function capture_muladd(ex::Expr, mod, @nospecialize(LHS) = nothing)
```
DilumAluthge, line 117: Same question as above: is there any risk that adding this `@nospecialize` will cause the runtime performance to be slower?

src/add_compute.jl
chriselrod: No. `add_pow!` expands literal powers into a product of squares while the macro expands.

timholy: I think not but for questions like these I'd defer to the experts.

src/constructors.jl
chriselrod: I guess this is a hint to inference that `q` will be an `Expr` from that point on.

timholy: The result type of `macroexpand` is not inferrable. This just allows the compiler to know that below this line, `q::Expr`.

src/graphs.jl
```diff
@@ -335,7 +335,7 @@ function LoopSet(mod::Symbol)
         Tuple{Int,Symbol}[],
         Tuple{Int,Int}[],
         Tuple{Int,Float64}[],
-        Int[],Int[],
+        Tuple{Int,NumberType}[],Tuple{Int,Symbol}[],
```
timholy, line 338: I've seen this kind of thing in my own code (e.g., https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/58), and I'm amazed that `convert(Vector{Tuple{Int,NumberType}}, ::Vector{Int})` doesn't error.

src/vectorizationbase_compat/contract_pass.jl
timholy: In this case I don't think the argument is even used.

timholy (commented):
I definitely don't have the answers here, so I'd ask @chriselrod to chime in.

src/add_compute.jl
```diff
@@ -222,7 +222,7 @@ function add_compute!(
     # instr = instruction(first(ex.args))::Symbol
     instr = instruction!(ls, first(ex.args))::Instruction
     args = @view(ex.args[2:end])
-    (instr.instr === :(^) && length(args) == 2 && (args[2] isa Number)) && return add_pow!(ls, var, args[1], args[2], elementbytes, position)
+    (instr.instr === :(^) && length(args) == 2 && (args[2] isa Number)) && return add_pow!(ls, var, args[1], args[2]::Number, elementbytes, position)
```
chriselrod, line 225: Would it be better to extract `args[2]` and then check that the extracted value is a number, rather than index twice?

src/add_loads.jl
```diff
@@ -2,13 +2,13 @@ function maybeaddref!(ls::LoopSet, op)
     ref = op.ref
     id = findfirst(r -> r == ref, ls.refs_aliasing_syms)
     # try to CSE
-    if isnothing(id)
+    if id === nothing
```
chriselrod, line 5: I should perhaps replace all `isnothing(x)` with `x === nothing`.

chriselrod (approved):
Fantastic as always! Ready to merge?

src/add_compute.jl
timholy: done

src/add_loads.jl
DilumAluthge: Yeah I think that's a good idea.

timholy:
For your interest, this was the workflow to find the inference issues:

```julia
using SnoopCompile, Cthulhu
tinf = @snoopi_deep include("some_workload.jl")
itrigs = inference_triggers(tinf)
itrig = itrigs[1]
ascend(itrig) # inspect and fix if needed (the first item in the chain is the callee that was runtime-dispatched or needed re-inferring, later ones are its callers)
itrig = itrigs[2]
ascend(itrig) # inspect and fix if needed...
```
The main gotcha is that the callers of a freshly-inferred MethodInstance are derived from a backtrace grabbed at the entrance to inference, and you only have a full MethodInstance for non-inlined methods. (Any `=>` you see in each row represents inlining.)

This won't be very rewarding on LoopVectorization now because it was already quite good and this eliminates most of the remaining "real" triggers; you just get a lot of noise from https://github.com/JuliaLang/julia/issues/38983. But in a package that has not yet received the kind of polish this one has, the results can be very informative.

src/vectorizationbase_compat/contract_pass.jl
chriselrod: It was once upon a time, but I changed where the `vfmadd` -> `vfmadd231` substitution takes place (it now happens during lowering, in `lower_compute!`).

It also no longer happens on Julia 1.6+. The reason for using `vfmadd231` was that LLVM used to not optimize many muladds well, choosing the wrong `vfmadd` version (overwriting the wrong registers), requiring the addition of move statements to copy register contents. So I used `asm call` to pick the correct version and avoid all the `vmovapd`s.
This has been fixed as of LLVM 11, which ships with Julia 1.6.

The commented out use of `LHS` is never getting un-commented, is what I'm trying to say.

Mind removing the argument altogether?

src/graphs.jl
chriselrod: Good catch!

Being empty saves Julia the need to convert an `Int` into a `Tuple{Int,Symbol}`, while "converting" the container (sans the contents) is trivial. Would be nice if this were easier to catch.
```julia
julia> convert(Vector{Tuple{Int,Symbol}}, Int[])
Tuple{Int64, Symbol}[]

julia> convert(Vector{Tuple{Int,Symbol}}, Int[3, 4])
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Tuple{Int64, Symbol}
Closest candidates are:
  convert(::Type{T}, ::T) where T<:Tuple at essentials.jl:317
  convert(::Type{T}, ::Tuple{Vararg{Any, N}}) where {N, T<:Tuple} at essentials.jl:318
  convert(::Type{T}, ::CartesianIndex) where T<:Tuple at multidimensional.jl:137
  ...
Stacktrace:
```

src/vectorizationbase_compat/contract_pass.jl
```diff
@@ -107,20 +107,20 @@ function recursive_muladd_search!(call, argv, cnmul::Bool = false, csub::Bool =
...
+function capture_muladd(ex::Expr, mod)
     call = Expr(:call, Symbol(""), Symbol(""), Symbol(""))
     found, nmul, sub = recursive_muladd_search!(call, ex.args)
     found || return ex
     # a, b, c = call.args[2], call.args[3], call.args[4]
     # call.args[2], call.args[3], call.args[4] = c, a, b
-    clobber = false#call.args[4] == LHS
+    clobber = false
     f = if nmul && sub
         clobber ? :vfnmsub231 : :vfnmsub
```
chriselrod, line 125: Could remove `clobber` as well, since it's trivially `false`

timholy: Good point, I didn't really look very carefully.

chriselrod:
> In retrospect they seem pretty obvious, but when profiling it never ceases to amaze me how often things only make sense in retrospect and are completely different from what I expected going in.

If someone hands you a needle and a piece of hay, it's also obvious which is which.
But if someone presents a haystack and asks you to find the needles, any tool to sort through it would be a lot of help.

Plus, it's not always obvious to those of us less familiar with Julia's internals!
E.g., when the compiler benefits from extra type hints.


> For your interest, this was the workflow to find the inference issues:

Thanks, I'm going to have to try that.
Much of that past polish is of course [to your credit](https://github.com/chriselrod/LoopVectorization.jl/pull/97), but `ascend` looks great and I know a few repos I expect to benefit a lot from this.


I intend to merge this (and VectorizationBase) once the final test passes, rerun benchmarks, and then issue a new release.
(I don't expect these changes to impact runtime performance, but I made a couple other changes since the last release that should.)

timholy:
Hah, I'd completely forgotten I'd added those precompiles! :laughing:

Issue #189, opened 2021-01-02
`@avx` is extremely slower than manually expanded loop


Roger-luo:
It's a bit to my surprise that in the following vector operation `@avx` is about 10x slower than a manual loop unroll

```julia
using BenchmarkTools
using LoopVectorization

function broutine_z!(st::Vector{Complex{T}}, ::Val{:Z}, locs) where T
    step_1 = 1 << (locs - 1)
    step_2 = 1 << locs

    @inbounds for j in 0:step_2:size(st, 1)-step_1
        for i in j:j+step_1-1
            ST1 = st[i+1]
            ST2 = -st[i+step_1+1]
        
            st[i+1] = ST1
            st[i+step_1+1] = ST2
        end
    end
    return st
end

function broutine_z_expanded!(st::Vector{Complex{T}}, ::Val{:Z}, locs) where T
    step_1 = 1 << (locs - 1)
    step_2 = 1 << locs

    @inbounds for j in 0:step_2:size(st, 1)-step_1
        for i in j:16:j+step_1-1
            Base.Cartesian.@nexprs 16 k->begin
                ST1 = st[i+k]
                ST2 = -st[i+step_1+k]
            
                st[i+k] = ST1
                st[i+step_1+k] = ST2
            end
        end
    end
    return st
end

function broutine_z_avx!(st::Vector{Complex{T}}, ::Val{:Z}, locs) where T
    mshift = locs
    mstride = 1 << mshift
    nstride = 1 << (mshift - 1)
    Mmax = size(st, 1)
    str = reinterpret(reshape, T, st)
    M = (Mmax-1) >>> mshift
    @avx for _m in 0:M
        m = _m << mshift
        for n in 1:nstride
            T1_re = str[1, m+n]
            T1_im = str[2, m+n]
            T2_re = -str[1, m+n+nstride]
            T2_im = -str[2, m+n+nstride]

            str[1, m+n] = T1_re
            str[2, m+n] = T1_im
            str[1, m+n+nstride] = T2_re
            str[2, m+n+nstride] = T2_im
        end
    end
    return st
end

st=rand(ComplexF64, 1 << 20);
broutine_z_avx!(copy(st), Val(:Z), 19) ≈ broutine_z!(copy(st), Val(:Z), 19)
broutine_z_expanded!(copy(st), Val(:Z), 19) ≈ broutine_z!(copy(st), Val(:Z), 19)

[...]

Issue #132, opened 2020-07-01
Strided Loops with @avx ?


LaurentPlagne:
Hi, I wonder if @avx can be used with strided loops like

```
function smooth_line(nrm1,j,i1,sl,rl,ih2,denom)
    @fastmath @inbounds @simd for i=i1:2:nrm1
            sl[i,j]=denom*(rl[i,j]+ih2*(sl[i,j-1]+sl[i-1,j]+sl[i+1,j]+sl[i,j+1]))
    end
end
```

It seems that, changing ```@fastmath @inbounds @simd``` by ```@avx``` does not lead to the same result

chriselrod:
Hi, thanks for the issue.
There are two problems with this loop currently:
1. I need to add support for strided loops. I'd have to think more about the best way to handle this. One possible solution is to forbid vectorizing them, but it would probably be better to allow that as a possibility. One idea is to treat the accessed arrays as having a dimension each to the stride ahead of the loop. I.e, `for n in 1:step:N; x[n]` would translate to `for n in 1:div(N,step); x[1,n]` where `x` is treated as having length `step` in the first dimension.
2. Loop carried dependencies aren't supported yet.

LaurentPlagne:
I Chris, thank for the reply.
1. I tried this solution and @avx produces (twice) slower code than @simd
```
@inline function smooth_line_avx(nrm1,j,i1,sl,rl,ih2,denom)
    ni2=div(nrm1,2)-1
    @avx for i2=0:ni2
        sl[i1+2i2,j]=denom*(rl[i1+2i2,j]+ih2*(sl[i1+2i2,j-1]+sl[i1+2i2-1,j]+sl[i1+2i2+1,j]+sl[i1+2i2,j+1]))
    end
end
```
2. I think there should be no dependencies in this loop (part of red-black gauss-seidel).

I think that the issue should be closed but, if it possible (I don't if it is), it would be nice to have a warning message saying that strided loops are not supported by @avx (or at least on the README).

BTW: are nested loops handled by @avx ?

chriselrod:
Mind sharing code I can copy and past to run the functions?

> I tried this solution and @avx produces (twice) slower code than @simd

Does `@simd` actually vectorize?
I assume it does not, and that not vectorizing is faster.

> I think there should be no dependency in this loop (part of red-black gauss-seidel).

The body is:
[...]

Issue #64, opened 2020-03-02
Vectorized gcd and lcm?


MasonProtter:
To support Matrices of `Rational` numbers in [Gaius.jl](https://github.com/MasonProtter/Gaius.jl) it would be really handy if we had vectorized greatest common denominator `gcd` and least common multiple `lcm` functions. Do you know of an easy way to support them in LoopVectorization?

chriselrod:
SIMDPirates.jl would probably be the place to add them. [Here](https://github.com/JuliaLang/julia/blob/master/base/intfuncs.jl) are the Julia definitions.

I [just added](https://github.com/chriselrod/SIMDPirates.jl/commit/047e753b746310708692c7a8dcab069e75faa294) a vectorized trailing zeros function (will probably only be vectorized with AVX512CD, but I found my [rand-log-from-integer](https://github.com/chriselrod/VectorizedRNG.jl/blob/master/src/special_approximations.jl#L259) function using leading zeros was faster on AVX2 anyway).

Those while loops are going to hurt. It'll have to repeat until every element of the vector meets the condition.

You're welcome to try and see if you can come up with a good implementation.

MasonProtter:
Thanks! Probably not a good candidate for vectorization then 😅

chriselrod:
Probably not, but "works" tends to be better than "does not work".
On the master branches of SIMDPirates and VectorizationBase:
```julia
julia> function Base.gcd(a::SVec{W,T}, b::SVec{W,T}) where {W, T <: Integer}
           #a == 0 && return abs(b)
           #b == 0 && return abs(a)
           za = trailing_zeros(a)
           zb = trailing_zeros(b)
           k = min(za, zb)
           u = unsigned(abs(a >> za))
           v = unsigned(abs(b >> zb))
           while true
               incomplete = u != v;
               incomplete == SIMDPirates.VectorizationBase.mask(Val(W), 0) && break
               ugv = u > v
               unew = vifelse(ugv, v, u)
               v = vifelse(ugv, u, v); u = unew
               vnew = v - u
               v = vifelse(incomplete, vnew >> trailing_zeros(vnew), v)
           end
           r = u << k
           # T(r) would throw InexactError; we want OverflowError instead
           #r > typemax(T) && __throw_gcd_overflow(a, b)
           r % T
       end

julia> v1 = SVec(Core.VecElement.((6,12, 24, 51)))
SVec{4,Int64}<6, 12, 24, 51>

[...]

PR #205, opened 2021-03-10
Use VecUnroll, add threading support

chriselrod:


DilumAluthge:
A few questions:
1. If I do `@avx for ... end` and I don't specify the value for `thread`, what is the default value that is used for `thread`?
2. Am I correct in saying that `@avxt for ... end` is exactly equivalent to doing `@avx thread=true for ... end`?
3. What happens if I do `@avxt thread=false for ... end`? Which value is used for `thread`? Similarly, what if I do `@avxt thread=3 for ... end`?

chriselrod:
I should definitely add docs that answer those questions.

> * If I do `@avx for ... end` and I don't specify the value for `thread`, what is the default value that is used for `thread`?

LoopVectorization picks. I thinking the picking algorithm and ramp-up as a function of loop sizes needs work.
It probably errs on the side of too few threads right now.

> * Am I correct in saying that `@avxt for ... end` is exactly equivalent to doing `@avx thread=true for ... end`?

Yes.

> * What happens if I do `@avxt thread=false for ... end`? Which value is used for `thread`? Similarly, what if I do `@avxt thread=3 for ... end`?

`@avxt thread=false for ... end` is equivalent to `@avx thread=false for ... end`.
Later arguments overrule earlier ones, and `@avxt thread=false for ... end` -> `@avx thread=true thread=false for ... end`
So the second (`false`) wins.

Same story with `thread=3`. However, `thread=number` sets the number as the maximum amount of threads to use. LoopVectorization can (currently) still decide to use less.
I may change that, because it certainly makes testing what the fastest number of threads is less convenient.

chriselrod:
fixes #132

Benchmark:
https://github.com/chriselrod/LoopVectorization.jl/issues/132#issuecomment-652575612

Performance, with the original code as from that comment:
```julia
julia> @show go(128)
norm(sl - sl_ref) = 0.0
(tsimd, tavx) = (1.904e-5, 1.5667e-5)
go(128) = (1.904e-5, 1.5667e-5)
(1.904e-5, 1.5667e-5)

julia> @show go(256)
norm(sl - sl_ref) = 0.0
(tsimd, tavx) = (7.349e-5, 6.4842e-5)
go(256) = (7.349e-5, 6.4842e-5)
(7.349e-5, 6.4842e-5)
```
Redefining `smooth_line_avx` like in the original non-working example (with seprange loops):
```julia
julia> @inline function smooth_line_avx(nrm1,j,i1,sl,rl,ih2,denom)
           @avx for i=i1:2:nrm1
                   sl[i,j]=denom*(rl[i,j]+ih2*(sl[i,j-1]+sl[i-1,j]+sl[i+1,j]+sl[i,j+1]))
           end
       end
smooth_line_avx (generic function with 1 method)

julia> @show go(128)
norm(sl - sl_ref) = 0.0
(tsimd, tavx) = (1.8695e-5, 1.6289e-5)
go(128) = (1.8695e-5, 1.6289e-5)
(1.8695e-5, 1.6289e-5)

julia> @show go(256)
norm(sl - sl_ref) = 0.0
(tsimd, tavx) = (7.1003e-5, 6.605e-5)
go(256) = (7.1003e-5, 6.605e-5)
(7.1003e-5, 6.605e-5)
```

chriselrod:
As of the commit that follows this comment (should push that commit within the next hour), fixes #189 :
```julia
julia> broutine_z_avx!(copy(st), Val(:Z), 19) ≈ broutine_z!(copy(st), Val(:Z), 19)
true

julia> @benchmark broutine_z_avx!(st, Val(:Z), 19) setup=(st=rand(ComplexF64, 1 << $n))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     140.593 μs (0.00% GC)
  median time:      216.166 μs (0.00% GC)
  mean time:        200.251 μs (0.00% GC)
  maximum time:     244.483 μs (0.00% GC)
  --------------
  samples:          3127
  evals/sample:     1

julia> @benchmark broutine_z!(st, Val(:Z), 19) setup=(st=rand(ComplexF64, 1 << $n))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     141.361 μs (0.00% GC)
  median time:      198.926 μs (0.00% GC)
  mean time:        200.677 μs (0.00% GC)
  maximum time:     246.956 μs (0.00% GC)
  --------------
  samples:          3122
  evals/sample:     1

julia> broutine_z_expanded!(copy(st), Val(:Z), 19) ≈ broutine_z!(copy(st), Val(:Z), 19)
true

julia> @benchmark broutine_z_expanded!(st, Val(:Z), 19) setup=(st=rand(ComplexF64, 1 << $n))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     139.300 μs (0.00% GC)
  median time:      200.017 μs (0.00% GC)
  mean time:        199.695 μs (0.00% GC)
  maximum time:     249.390 μs (0.00% GC)
  --------------
  samples:          2947
  evals/sample:     1
```

chriselrod:
The bump in compat to require VectorizationBase 0.19.6
fixes #64 
fixes #206

chriselrod:
All tests pass! I'll merge soon.

chriselrod:
Assuming all tests pass again, I'll merge and issue a new release.

Then PRs for Gaius.jl, Octavian.jl, and BLASBenchmarksCPU.jl.

PR #210, opened 2021-03-17
tweak package import in file loaded by Requires

KristofferC:
If one bundle this package into a sysimage, it is quite easy to run into problems because the package will at Julia start time try to load `VectorizationBase` which will cause a full package lookup via the project and manifest of the currently active project. If one does not have a project active such that this is possible when Julia starts, it will error. There is actually no need for this, we can just use the VectorizationBase that is already loaded in the package.


KristofferC:
Not sure what is up with CI. Can't even see any of the logs.

chriselrod:
GitHub disabled all GitHub actions for baby of my repos.

PR #212, opened 2021-03-20
Update README.md

GiggleLiu:
add TropicalGEMM to the list of packages using LoopVectorization.

Is this the correct place to go? or Octavian is better?


chriselrod:
> Is this the correct place to go? or Octavian is better?

Both?

The documentation and README here are pretty out of date overall.

Not sure what the message is, but partly because both libraries are still fairly immature.
I'll start worrying more about presentation once LoopVectorization supports triangular loops, dependency analysis, and cache tiling optimizations; that's the point where I'll start being more interested in messaging.
Realistically though, I'll still procrastinate on that until there's autodiff support and the frontend uses AbstractInterpreter too. And by then, I'll almost certainly have found new things to prioritize ahead of working on documentation.

PR #240, opened 2021-04-16
add another assumption of at-avx to the README.md

ranocha:
I encountered an issue when using nested loops with multiple loops at the same level. This comment might help other users of this great package when running into such an issue. Did I understand the comment https://github.com/JuliaSIMD/LoopVectorization.jl/issues/230#issuecomment-810632972 correctly?


chriselrod:
Thanks!
Yes, I'll add support for that in the rewrite.

It'll be able to do optimizations like fusing the loops.

Issue #238, opened 2021-04-15
segfault with static sizes and related issues


ranocha:
I get the following error on Julia v1.6.0 using LoopVectorization v0.12.12, StrideArrays v0.1.5, StrideArraysCore v0.1.5.
```julia
julia> using LoopVectorization, StrideArrays

julia> function issue!(output, matrix, input)

         @avx inline=true for j in axes(output, 3), i in axes(output, 2), v in axes(output, 1)
           res = zero(eltype(output))
           for ii in axes(matrix, 2)
             res += matrix[i, ii] * input[v, ii, j]
           end
           output[v, i, j] = res
         end

         return nothing
       end
issue! (generic function with 1 method)

julia> input = randn(3, 4, 4); matrix = randn(7, 4); output = StrideArray(undef, StaticInt(3), StaticInt(7), StaticInt(4)); output_nonstatic = StrideArray(undef, 3, 7, 4);

julia> issue!(output_nonstatic, matrix, input) # that's fine

julia> issue!(output, matrix, input)
ERROR: MethodError: no method matching isless(::Val{12}, ::Int64)
Closest candidates are:
  isless(::Any, ::Missing) at missing.jl:88
  isless(::Missing, ::Any) at missing.jl:87
  isless(::AbstractFloat, ::Real) at operators.jl:169
  ...
Stacktrace:
  [1] <(x::Val{12}, y::Int64)
    @ Base ./operators.jl:279
  [2] <=(x::Val{12}, y::Int64)
    @ Base ./operators.jl:328
  [3] mask_type(W::Val{12})
    @ VectorizationBase ~/.julia/packages/VectorizationBase/czbgP/src/early_definitions.jl:113
  [4] EVLMask
    @ ~/.julia/packages/VectorizationBase/czbgP/src/VectorizationBase.jl:119 [inlined]
  [5] EVLMask
    @ ~/.julia/packages/VectorizationBase/czbgP/src/VectorizationBase.jl:124 [inlined]
  [6] macro expansion
    @ ~/.julia/packages/VectorizationBase/czbgP/src/llvm_intrin/masks.jl:617 [inlined]
  [7] mask
    @ ~/.julia/packages/VectorizationBase/czbgP/src/llvm_intrin/masks.jl:617 [inlined]
  [8] macro expansion
    @ ~/.julia/packages/VectorizationBase/czbgP/src/llvm_intrin/masks.jl:253 [inlined]
  [9] _mask
[...]

PR #247, opened 2021-04-24
add Trixi to users of LoopVectorization

ranocha:
We're already using LoopVectorization to get some really nice improvements in our hyperbolic PDE framework. We will use LoopVectorization even more and also other great packages of yours like CheapThreads and StrideArrays once #238 (and maybe https://github.com/JuliaSIMD/VectorizationBase.jl/issues/53) are resolved.


chriselrod:
Thanks! 
#238 should be fixed on master. I was planning on going over issues and testing/closing them this weekend.

I'll also look at JuliaSIMD/VectorizationBase.jl#53 and either define a bunch of fallback methods, or figure out why dispatch is hitting them and address that.

ranocha:
Thanks a lot!

PR #270, opened 2021-05-27
parallel tests

ranocha:
I observed that the CI tests take quite a while for LoopVectorization. I guess this can be annoying since you have to wait for several hours to see whether PRs pass all tests. Here, I propose to split the tests into three parts which can be run in parallel to reduce wall clock time for CI.

If this is acceptable, the required CI tests need to be updated.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,18 @@
+# https://docs.codecov.io/docs/codecovyml-reference
+
+# There are
+#  3 [number of parts] * 3 [number of OS] = 9
+# parallel jobs in ci.yml
+codecov:
+  notify:
+    after_n_builds: 9
+comment:
+  after_n_builds: 9
+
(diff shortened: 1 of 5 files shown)


chriselrod (approved):
Looks good to me.
Tests fail due to timing out on occasion (e.g., Windows on Julia nightly timed out on the last commit to master).

While local tests are much faster (assuming you aren't recording coverage).
Eventually, would be nice to make tests run in parallel locally as well.
Could be done via `run`inning new Julia sessions when `group = "all"`, and storing results to `.gitignore`d files.

chriselrod:
> If this is acceptable, the required CI tests need to be updated.

Yeah. I'll merge when tests pass, and then update the required tests.

chriselrod:
Excellent, thank you!

PR #273, opened 2021-05-28
Add logo to readme

MasonProtter:
Just in case you wanted the new logo in the README.md, I figured I'd add it. No pressure to accept if you think it's taking up too much space. Can also try and change the size or location


chriselrod:
Thanks!

PR #275, opened 2021-05-29
add link to the repo to the docs

ranocha:
Sometimes, people will find the docs via Google etc. at first. Then, it canbe nice to have a link back to the repo on GitHub.


chriselrod (approved):
Great, thanks!

PR #306, opened 2021-07-25
Interpret some constant expressions, fixes #305

### LINKED ISSUE #305: Segfault in complex finite difference stencil in multithreaded
I made a stencil generator for a 3D finite difference model that generates a complex stencil on a staggered grid. The usage of `@tturbo` results in a segfault in multithreaded mode, although the single thread performance is excellent. The code can be found on GitHub [here](https://github.com/microhh/stencilbuilder/blob/master/julia/stencil_julia_4th.jl)

Since this is my first Julia experience, I would like to learn whether this is a bug, or whether I ask too much from `LoopVectorization`.

chriselrod:
I'll add more accurate determination of register pressure in another commit, probably to this same PR. The associated issue for that is [this comment](https://julialang.zulipchat.com/#narrow/stream/137791-general/topic/Working.20with.20Complex.20Numbers.20in.20LoopVectorization/near/247055705) on Zulip.

For now, it just contains the interpretation of some constant expressions, which can sometimes drastically reduce the amount of arguments passed to `_turbo_!`.

### DIFF
--- a/.github/workflows/ci-julia-nightly.yml
+++ b/.github/workflows/ci-julia-nightly.yml
@@ -39,6 +39,8 @@ jobs:
           - part4
           - part5
           - part6
+          - part7
+          - part8
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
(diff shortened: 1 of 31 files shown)


chriselrod:
Oh, and add `vfmaddsub` support, which I apparently haven't actually released in VectorizationBase yet.

PR #315, opened 2021-08-04
Fix typo; remove outdated naming parenthetical

palday:
The `named MKL` bit is no longer accurate: the plot just says "Julia" (also vanilla Julia is on OpenBLAS).


chriselrod:
I don't recall if I was on MKL or OpenBLAS for Julia last time I ran the benchmarks.
With Julia 1.7 and up, it's easy to `using MKL`, but the last time I ran the benchmarks definitely predated this.

chriselrod:
There are some test failures on master because of changes to VectorizationBase.
I'll merge this once I've resolved them.

PR #316, opened 2021-08-13
Delete `convert(::Type{Any}, ::UpperBoundedInteger)`

timholy:
This method isn't necessary (it just replicates the fallback),
and it triggers ~600 invalidations.

CC @ChrisRackauckas

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LoopVectorization"
 uuid = "bdcacae8-1622-11e9-2a5c-532679323890"
 authors = ["Chris Elrod <elrodc@gmail.com>"]
-version = "0.12.63"
+version = "0.12.64"
 
 [deps]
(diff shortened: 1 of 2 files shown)


chriselrod:
Ouch. Mind bumping the version?

I haven't had the time to go over your comment [here](https://github.com/JuliaSIMD/TriangularSolve.jl/pull/8#issuecomment-895398355), but did check and see what looked like a lot of invalidation in the LV ecosystem.

My plan is still to rewrite the current library when I have time, so I'm also not sure how much effort is worth spending optimizing the old library. But low hanging fruit like this seem well worth it if they shave off compile times.

Would be nice to get @ChrisRackauckas 22s -> 2.5s down to <= 2.2 so he can claim 10x faster.

ChrisRackauckas:
>Would be nice to get @ChrisRackauckas 22s -> 2.5s down to <= 2.2 so he can claim 10x faster.

Yes, not hitting that mark definitely made the marketing harder.

PR #322, opened 2021-08-23
Set timeout-minutes

rikhuijzer:
Without this, GitHub Actions runs for 6 hours when nightly is ending up in some infinite loop again (like currently seems to be the case).


rikhuijzer:
Wow. So apparently runs take about 30 minutes on 1 hour and 10 minutes on another (https://github.com/JuliaSIMD/LoopVectorization.jl/runs/3401454253?check_suite_focus=true). Still, they do pass, so this PR is probably a bad idea. Feel free to close.

chriselrod:
Yeah, before splitting up the test suite it'd take several hours to run and risk hitting CI's time limit.

PR #341, opened 2021-09-25
Add tests that mimic array-of-struct RGB

timholy:
This is the default storage in JuliaImages. This ensures that methods
based on LoopVectorization will continue to work (an insurance policy
against future changes), and also makes it easy to check performance
(manually, since CI is bad for performance tests).

Currently, the performance on the RGB does not match the "in principle" 3x the corresponding grayscale. For a 3d output image of size 11x11x11, here are the timings I'm getting for filtering along a single dimension `d`:

| d | t grayscale (μs) | t RGB (μs) | Expected based on grayscale | `@simd` only |
| - | ----------:| -----------:| -----:| ----:|
| 1 | 4.5 | 27.5 | 13.5 | 42.2 |
| 2 | 1.8 | 14.1 | 5.4 | 48.5 |
| 3 | 1.1 | 5.2 | 3.3 | 48.8 |

Not sure if it's even fixable (meaning, we should switch to struct-of-arrays representation), but I wanted to know the answer and thought it would be better to document it. Certainly, especially for `d>1` the performance gains relative to plain `@simd` (even with which LLVM fails to vectorize) are impressive.

Note that I constructed `Rpre` with a static range for its first range, hoping it would help.

### DIFF
--- a/test/miscellaneous.jl
+++ b/test/miscellaneous.jl
@@ -1,7 +1,6 @@
 using LoopVectorization
 using LinearAlgebra
 using OffsetArrays
-using Static
 using Test
 
 if !isdefined(@__MODULE__, Symbol("@_avx"))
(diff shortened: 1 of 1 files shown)


chriselrod:
Thanks as always for the tests!

> also makes it easy to check performance
(manually, since CI is bad for performance tests).
> 
> Currently, the performance on the RGB does not match the "in principle" 3x the corresponding grayscale. For a 3d output image of size 11x11x11, here are the timings I'm getting for filtering along a single dimension d:

I'll take a look. In theory, some CPUs should be able to get close to the theoretic value.

> Note that I constructed `Rpre` with a static range for its first range, hoping it would help.

It should help. Knowing the static size of any dimensions that are that small should help. LV assumes a length of around 1024 for dynamic dimensions, so when an axis is very small (particularly less than a few vector widths) this can be innaccuarate.

Also:
```julia
julia> using ArrayInterface

julia> x = [(rand(),rand(),rand()) for i in 1:11, j in 1:11];

julia> xr = reinterpret(reshape, Float64, x);

julia> ArrayInterface.axes(xr)
(Base.OneTo(3), Base.OneTo(11), Base.OneTo(11))

julia> ArrayInterface.axes(xr,1)
static(1):static(3)
```
The first one returns `Base.OneTo(3)` because there is a [TODO](https://github.com/JuliaArrays/ArrayInterface.jl/blob/ad11cc1d4572c9679b3a99c8ebaf194035ec9ab9/src/axes.jl#L177) to implement an ArrrayInterface version. Seems like it should just be a `map`.

PR #373, opened 2021-12-27
Drop Requires

chriselrod:
@ChrisRackauckas this increases loading time from 2 to 3 seconds on my computer, but I'll merge anyway if you think it's worth it. Odds are most people are using `ForwardDiff`, `SpecialFunctions`, and `ChainRulesCore` anyway, at least indirectly.


ChrisRackauckas:
Is that the change in load time when including the `using` of the 3 packages being Requires'd??

chriselrod:
> Is that the change in load time when including the `using` of the 3 packages being Requires'd??

Using all four packages, the time went down from 3.2 to 3.1 seconds.

chriselrod:
The margin is similar for `using OrdinaryDiffEq`. On master, I get 6.79 s vs 6.65 s on this PR.

OrdinaryDiffEq is probably used at least 10x more often, so probably an overall win here?

chriselrod:
The margin is similar for `using OrdinaryDiffEq`. On master, I get 6.79 s vs 6.65 s on this PR.

OrdinaryDiffEq is probably used at least 10x more often, so probably an overall win here?

ChrisRackauckas:
Yeah, I'd be curious to figure out how it's still at 3 seconds though.

chriselrod:
1.5s comes from VectorizationBase.

0.15s can be shaved off by removing the precompiles from LoopVectorization.

I don't think it's worth spending too much time on this. The long term plan is for us to all move away from LoopVectorization and the entire JuliaSIMD ecosystem, so every effort spent here will be for the short term only, and I don't think that's really worth our time.

But easy wins are always nice.

ma-sadeghi:
> I don't think it's worth spending too much time on this. The long term plan is for us to all move away from LoopVectorization and the entire JuliaSIMD ecosystem, so every effort spent here will be for the short term only, and I don't think that's really worth our time.

Just curious, is the JuliaSIMD ecosystem (including LoopVectorization) going to be built-in in Julia itself?

chriselrod:
> > I don't think it's worth spending too much time on this. The long term plan is for us to all move away from LoopVectorization and the entire JuliaSIMD ecosystem, so every effort spent here will be for the short term only, and I don't think that's really worth our time.
> 
> Just curious, is the JuliaSIMD ecosystem (including LoopVectorization) going to be built-in in Julia itself?

I'm rewriting it to work at the LLVM level.
It'll no longer use any of the existing JuliaSIMD ecosystem, and fix many of the associated problems, the major highlights being:
1. lower latency
2. it'll support loop carried dependencies
3. it'll support non-rectangular loop nests
4. it'll support multiple loops at the same level of a nest

It'll exist as a separate package for at least a while.
But depending on how things go, it is conceivable that people will be interested in getting it into base Julia and dropping OpenBLAS. But I do not expect that to happen any time soon, and would also expect a lot of opposition to a change that radical.

chriselrod:
fixes #374 

Not sure if commenting on the PR is enough for the issue to be closed automatically once the PR merges, but I'll find out.

ma-sadeghi:
> Not sure if commenting on the PR is enough for the issue to be closed automatically once the PR merges, but I'll find out.

@chriselrod I think you need to add "fixes #374" to the PR body (i.e., edit the first comment in the PR) for it to close the issue.

PR #382, opened 2022-02-09
Static compat includes "0.5"

JeffreySarnoff:


chriselrod:
LoopVectorization is not compatible with Static 0.5.

JeffreySarnoff:
ok

PR #389, opened 2022-03-30
Include discussion of performance vs. accuracy tradeoffs and comparison to fastmath in docs

danielwe:
This draft PR is based on a recent slack thread about the arithmetic transformations and function substitutions performed by `@turbo` and how they affect both accuracy and performance, as well as how they compare to the transformations performed by `@fastmath`. Thanks to @chriselrod and @oscardssmith for the incredibly helpful discussion. It would be great to preserve these insights and make them easy to find for future users. For now, I've copied the discussion verbatim over to a separate page in the docs. Aiming to come back and edit this into prose soon, but maintainers are also welcome to edit freely.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -20,6 +20,7 @@ Pages = [
     "examples/special_functions.md",
     "examples/sum_of_squared_error.md",
     "vectorized_convenience_functions.md",
+    "performance_accuracy_fastmath.md",
     "future_work.md",
     "devdocs/overview.md",
     "devdocs/loopset_structure.md",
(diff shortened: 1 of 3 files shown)


chriselrod:
Are you planning on getting back to this?
If you don't want to edit the content, you could just make a comment that this was a slack discussion, and that anyone reading it is welcome to make a PR to turn the conversation into a more concise explanation.

danielwe:
Yeah, realistically it would probably be a while before I'd get around to editing this. How does it look now?

chriselrod:
Yeah, I think this is fine, thanks!

PR #394, opened 2022-04-09
Replace 0x0000000000000000 with zero(UInt) for 32-bit systems

brenhinkeller:
May avoid problems on 32-bit systems (if anyone is using Julia on those anyways?)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LoopVectorization"
 uuid = "bdcacae8-1622-11e9-2a5c-532679323890"
 authors = ["Chris Elrod <elrodc@gmail.com>"]
-version = "0.12.105"
+version = "0.12.106"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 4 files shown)


chriselrod (approved):
I won't auto-merge, as Macs have been failing on master.
So I'll wait for (other) tests to pass and then merge manually.

chriselrod:
Huh, actually, there is a red x on the last commit into master, yet I couldn't find the test that actually failed.

brenhinkeller:
Oh I think the red x might actually just be from the codecov?

brenhinkeller:
Yeah I think it's just this -0.4% from statistical noise

Screen Shot 2022-04-08 at 11 38 54 PM

I copy-pasted some stuff into my codecov.yml that seems to stop that from happening (e.g. https://github.com/brenhinkeller/NaNStatistics.jl/blob/main/codecov.yml -- though not sure if it's the "threshold" or "target" value or both that actually does the trick)

chriselrod:
Looks like all nightly tests have passed.

brenhinkeller:
Tests look good on nightly on my mac and 32-bit tests are passing on NaNStatistics so I'll call it success! Nightly may just be a bit flakey lately

Issue #395, opened 2022-04-09
`@avx` for loop whose body is not a `block` causes  `StackOverflowError`


wangl-cc:
I'm trying to generate loops with macro and use `@avx` to accelerate it. For examples,
```julia
macro gen_loop_avx_1(ex)
    sym, ind = ex.args
    loop_body = :(ret[$ind] = $sym[$ind])
    loop = Expr(:for, :($ind = axes($sym, 1)), loop_body)
    return esc(:(@avx $loop))
end
```
this macro will generate a loop like this with `@macroexpand1 @gen_loop_avx_1 V[i]`
```julia
@avx for i = axes(V, 1)
    ret[i] = V[i]
end
```
However, when expand macros with `@macroexpand`, I receives a `StackOverflowError`.

A solution is make `loop_body` becomes a `block`:
```Julia
macro gen_loop_avx_2(ex)
    sym, ind = ex.args
    loop_body = Expr(:block, :(ret[$ind] = $sym[$ind]))
    loop = Expr(:for, :($ind = axes($sym, 1)), loop_body)
    return esc(:(@avx $loop))
end
```
this macro will generate the similar loop but its body is a block and works well with `@avx`.

Then, when I check my original code, the loop body is defined as a `block` until
`MacroTools.flatten` remove it, like this:
```Julia
macro gen_loop_avx_3(ex)
    sym, ind = ex.args
    loop_body = Expr(:block, :(ret[$ind] = $sym[$ind]))
    loop = Expr(:for, :($ind = axes($sym, 1)), loop_body)
    return esc(:(@avx $(MacroTools.flatten(loop))))
end
```
which will cause `StackOverflowError` again.

In my opinion, the `@avx` checking the loop body if it is a block may be helpful.

chriselrod:
The simplest fix may be to check if it is a block, and if not wrap it with one:
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/77d0efb00184305fd8c4b9cfe1b5c5d28d9c108f/src/constructors.jl#L161-L162
Feel free to make a PR.

wangl-cc:
@chriselrod Hi, I'm working on it, where should tests about this issue be put?

chriselrod:
Feel free to create a new file called "parsing_inputs.jl" or something.
For where to put the `include`, put it in whichever group is fastest:
[...]

PR #396, opened 2022-04-09
fix: #395, check if loop body is a block and warp it if not

wangl-cc:
There are four macros which will generate loop whose body is not a block. The `gen_loop_issue395` and `gen_A_mul_B` macro cause `StackOverflowError` in current master while the `gen_single_loop` and `gen_nest_loop` cause `LoadError: TypeError: in typeassert, expected Int64, got a value of type Static.StaticInt{1}`. I have no idea about these errors, but warp those loop body as a block that fixes all of these error.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -159,7 +159,7 @@ end
 # check if the body of loop is a block, if not convert it to a block issue#395
 function check_loopbody!(q)
   if q isa Expr && q.head == :for
-    if q.args[2].head != :block
+    if Meta.isexpr(q.args[2], :block)
       q.args[2] = Expr(:block, q.args[2])
     else
       for arg in q.args[2].args
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -156,9 +156,24 @@ function process_args(
   end
   inline, check_empty, u₁, u₂, v, threads, warncheckarg
 end
+# check if the body of loop is a block, if not convert it to a block issue#395
+function check_loopbody!(q)
+  if q isa Expr && q.head == :for
+    if q.args[2].head != :block
+      q.args[2] = Expr(:block, q.args[2])
+    else
+      for arg in q.args[2].args
+        check_loopbody!(arg) # check recursively for inner loop
+      end
+    end
+  end
+  return q
+end
+
 function turbo_macro(mod, src, q, args...)
   q = macroexpand(mod, q)
   if q.head === :for
+    check_loopbody!(q)
     ls = LoopSet(q, mod)
     inline, check_empty, u₁, u₂, v, threads, warncheckarg = process_args(args)
     esc(setup_call(ls, q, src, inline, check_empty, u₁, u₂, v, threads, warncheckarg))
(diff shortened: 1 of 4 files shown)


src/constructors.jl
```diff
@@ -156,9 +156,24 @@ function process_args(
   end
   inline, check_empty, u₁, u₂, v, threads, warncheckarg
 end
+# check if the body of loop is a block, if not convert it to a block issue#395
+function check_loopbody!(q)
+  if q isa Expr && q.head == :for
+    if q.args[2].head != :block
```
chriselrod, line 162:
```suggestion
    if q.args[2].head !== :block
```
Testing identity is likely a little faster, because it is a non-overloadable builtin, meaning there's no dispatch and it's guaranteed to return `Bool`, while type unstable `==` and `!=` can return anything. 
This isn't a big deal here, as `if` only accepts `Bool`, meaning we effectively have a type assert on the return anyway. 
So something to keep in mind when working on `Expr` more generally, as their `.args` are `Vector{Any}`. 

Otherwise, a more succinct check:
```
    if Meta.isexpr(q.args[2], :block)
```

test/parsing_inputs.jl
```diff
@@ -0,0 +1,50 @@
+using LoopVectorization, Test, ArrayInterface
+
+# macros for generate loops whose body is not a block
+macro gen_loop_issue395(ex)
+  sym, ind = ex.args
+  loop_body = :(ret[$ind] = $sym[$ind])
+  loop = Expr(:for, :($ind = axes($sym, 1)), loop_body)
+  return esc(:(@avx $loop))
```
chriselrod, line 8: `@turbo` is preferred for forward-compatibility. 
0.13 will deprecate `@avx`, and 0.14 will remove it. 

Although, there are no plans to actually release a 0.13 anytime soon. 
I don't see the point in breaking anything.

chriselrod (commented):
Mind bumping the patch version?

wangl-cc:
@chriselrod Should I squash?

src/constructors.jl
wangl-cc: > So something to keep in mind when working on Expr more generally, as their .args are Vector{Any}.

Thanks for you suggestion.

chriselrod:
> @chriselrod Should I squash?

I enabled auto-merge squash, so it'll automatically merge once tests pass.
I'll have to re-enable it if you push any new commits, though.

wangl-cc:
I don't know reason, it seams that `Meta.isexpr` break the fix. The `!=` works see https://github.com/wangl-cc/LoopVectorization.jl/runs/5955789898?check_suite_focus=true.

chriselrod:
> ~I don't know reason, it seams that `Meta.isexpr` break the fix. The `!=` works see https://github.com/wangl-cc/LoopVectorization.jl/runs/5955789898?check_suite_focus=true.~
> 
> I'm sorry, I forget a `!`.

My bad, I think I didn't have it in my comment.

chriselrod:
Thanks!

Issue #393, opened 2022-04-07
Can't turbo or tturbo enumerate blocks


OmegaLambda1998:
Both of the following code blocks produce `ERROR: LoadError: TypeError: in typeassert, expected Symbol, got a value of type Expr`
```
@turbo for (i, n) in enumerate(1:100)
begin end
end
```

```
@tturbo for (i, n) in enumerate(1:100)
begin end
end
```
wrapping `enumerate` in `collect` does not resolve this issue

chriselrod:
I'm focusing most of my free/SIMD time on the rewrite, which will not have this problem.
So I've largely stopped putting time into LoopVectorization.jl issues/bugs.
If I didn't, I'd never have the time to actually work on the replacement.

It'll likely still be a long time before it is ready to replace LoopVectorization (e.g. a year or so is my current estimate).

So all that said, this would be a good first issue if someone wants to try adding support for this themselves.

The basic approach would be to basically replace
```julia
for (i,n) in enumerate(x)

end
```
with
```julia
for i in 1:length(x)
    n = x[i+firstindex(x)-1]
end
```
If anyone is interested, I can provide a more detailed explanation of ideas for how to implement that in the code and answer any questions.

chriselrod:
Fixed in https://github.com/JuliaSIMD/LoopVectorization.jl/pull/396

wangl-cc:
Hi, the PR #396 doesn't seems to fix this issue. @chriselrod

chriselrod:
> Hi, the PR #396 doesn't seems to fix this issue. @chriselrod

I closed the wrong issue by mistake.

chriselrod:
Closed by #397.

PR #397, opened 2022-04-10
feat: replace enumerate

wangl-cc:
I'm working on the issue #393 by replacing enumerate which can be integrated in the check of loop body. However, I have some difficulties.
First, turbo for a expression instead of a symbol may cause `LoadError` like:
```Julia
@turbo for i in 1:length(1:10)
    n = (1:10)[i+firstindex(1:10)-1]
end
```
may cause `ERROR: LoadError: "Indexing into the following expression was not recognized: 1:10"`. This error is not related to this PR and not common, feel free to ignore it, if this is not allowed.
Second, for
```Julia
@turbo for ix in enumerate(A)
    A[ix[1]] = ix[1] + ix[2]
end
```
where the index `i` and variable `x` are not unboxed. I tried to generate symbols firstly and regroup them as a tuple like this
```Julia
for ix_ind = Base.OneTo(length(A))
    ix_var = A[ix_ind + firstindex(A) - 1]
    ix = (ix_ind, ix_var)
    A[ix[1]] = ix[1] + ix[2]
end
```
where the expression `(ix_ind, ix_var)` can't not be recognized. Maybe this form should not be allowed.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -157,23 +157,71 @@ function process_args(
   inline, check_empty, u₁, u₂, v, threads, warncheckarg
 end
 # check if the body of loop is a block, if not convert it to a block issue#395
-function check_loopbody!(q)
-  if q isa Expr && q.head == :for
+# and check if the range of loop is an enumerate, if it is replace it, issue#393
+function check_inputs!(q)
+  if Meta.isexpr(q, :for)
     if !Meta.isexpr(q.args[2], :block)
       q.args[2] = Expr(:block, q.args[2])
-    else
+      replace_enumerate!(q) # must after warp block
+    else # maybe inner loops in block
+      replace_enumerate!(q)
       for arg in q.args[2].args
-        check_loopbody!(arg) # check recursively for inner loop
+        check_inputs!(arg) # check recursively for inner loop
       end
     end
   end
   return q
 end
+function replace_enumerate!(q)
+  looprange = q.args[1]
+  if Meta.isexpr(looprange, :block)
+    for i in 1:length(looprange.args)
+      convert_single_enumerate!(q, i)
+    end
+  else
+    convert_single_enumerate!(q)
+  end
+  return q
+end
+function convert_single_enumerate!(q, i=nothing)
+  if isnothing(i) # not nest loop
+    looprange, body = q.args[1], q.args[2]
+  else # nest loop
+    looprange, body = q.args[1].args[i], q.args[2]
+  end
+  @assert Meta.isexpr(looprange, :(=), 2)
+  itersyms, r = looprange.args
+  if Meta.isexpr(r, :call, 2) && r.args[1] == :enumerate
+    iter = r.args[2]
+    if Meta.isexpr(itersyms, :tuple, 2)
+      indsym, varsym = itersyms.args[1]::Symbol, itersyms.args[2]::Symbol
+      _replace_looprange!(q, i, indsym, iter)
+      pushfirst!(body.args, :($varsym = $iter[$indsym + firstindex($iter) - 1]))
+    elseif Meta.isexpr(itersyms, :tuple, 1) # like `for (i,) in enumerate(...)`
+      indsym = itersyms.args[1]::Symbol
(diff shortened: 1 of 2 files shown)


chriselrod:
> First, turbo for a expression instead of a symbol may cause `LoadError` like:

It's not "not allowed" so much as "not supported yet".
I guess it's not common for a human to write code like that, with the range directly inserted and indexed into.
A human would just offset `i` instead of using `i` to index into a literal range.

> where the index i and variable x are not unboxed. I tried to generate symbols firstly and regroup them as a tuple like this

Ideally, you could search for all `ix[1]`s and `ix[2]`s and replace them with `ix_ind` and `ix_var`, respectively:
```julia
for ix_ind = Base.OneTo(length(A))
    ix_var = A[ix_ind + firstindex(A) - 1]
    A[ix_ind] = ix_ind + ix_var
end
```
However, you could also disallow it for now, because what if someone was doing something other than indexing into `ix`?

wangl-cc:
> I guess it's not common for a human to write code like that, with the range directly inserted and indexed into.
> A human would just offset `i` instead of using `i` to index into a literal range.

A human may not write code like this, but replacing enumerate may generated this code.
A enumerate for a literal range like this
```Julia
for (i, n) in enumerate(1:2)
end
```
would be replaced like this:
```Julia
for i = Base.OneTo(length(1:2))
  n = (1:2)[(i + firstindex(1:2)) - 1]
end
```
One possible solution is naming the literal range in loop like this
```Julia
for i = Base.OneTo(length(1:2))
  r = 1:2
  n = r[(i + firstindex(r)) - 1]
end
```
but turbo on it shows `r` is not defined.
Another solution is naming it out of loop, but I have no idea how to do this.

Maybe create a block warp the `setup_call` and add naming before it at here?
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/master/src/constructors.jl#L178-L179

chriselrod:
> Another solution is naming it out of loop, but I have no idea how to do this.

I would definitely name it out of the loop.

chriselrod:
It would probably be easier to do this within the `LoopSet` constructor then.

src/constructors.jl
```diff
@@ -157,23 +157,71 @@ function process_args(
...
+end
+_replace_looprange!(q, ::Nothing, indsym, iter) = q.args[1] = :($indsym = Base.OneTo(length($iter)))
+_replace_looprange!(q, i::Int, indsym, iter) = q.args[1].args[i] = :($indsym = Base.OneTo(length($iter)))
 
 function turbo_macro(mod, src, q, args...)
   q = macroexpand(mod, q)
   if q.head === :for
-    check_loopbody!(q)
+    check_inputs!(q)
     ls = LoopSet(q, mod)
```
chriselrod, line 225: The simplest way might be to replace this line with `ls = LoopSet(mod)`, and move it above `check_inputs!` so you have access to the `prepreamble` for pushing the hoisted expression.
Then, after you've checked inputs, you can finish the rest of the constructor, like here:
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/d0771404645b579ec6002684d0ffed66699f4bd1/src/constructors.jl#L81-L85

chriselrod: I.e., basically
```julia
 ls = LoopSet(mod)
 check_inputs!(q, ls.prepreamble)
 contract_pass!(q)
 copyto!(ls, q) 
 resize!(ls.loop_order, num_loops(ls)) 
 ls
```

wangl-cc: Thanks a lot.

src/constructors.jl
```diff
@@ -78,8 +78,9 @@ end
 
 
 function LoopSet(q::Expr, mod::Symbol = :Main)
-  contract_pass!(q)
   ls = LoopSet(mod)
+  check_inputs!(q, ls.prepreamble)
```
chriselrod, line 82: Yes, moving `check_inputs!` in here is better.

chriselrod (approved):
Looks good to me, thanks!

PR #412, opened 2022-06-09
Fix contiguous_axis

Tokazama:
Locally, this gets everything working with the current master branch of ArrayInterface

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -29,9 +29,9 @@ UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 VectorizationBase = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
 
 [compat]
-ArrayInterface = "6.0.1 - 6.0.11"
+ArrayInterface = "6"
 ArrayInterfaceCore = "0.1.5"
-ArrayInterfaceOffsetArrays = "0.1.2"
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -29,9 +29,9 @@ UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 VectorizationBase = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
 
 [compat]
-ArrayInterface = "6.0.1 - 6.0.11"
+ArrayInterface = "6"
 ArrayInterfaceCore = "0.1.5"
-ArrayInterfaceOffsetArrays = "0.1.2"
```
chriselrod, line 34: Why drop the 0.1.2 requirement?

Tokazama: Because the most recent changes fix the issues caused by the forwarding trait

chriselrod: This was a lower bound.
Because of:
https://github.com/JuliaArrays/ArrayInterface.jl/issues/272
I cannot easily double check what actually changed in that version.

I'd be more comfortable leaving the lower bound of 0.1.2.

Tokazama: Oh I didn't explicitly test the lower versions. I was just trying to get everything working with master going forward

PR #416, opened 2022-06-17
Update Project.toml

Tokazama:


chriselrod:
I clicked on a couple of these, and it was using Static 0.6.6.

Tokazama:
Sorry. I should have commented that I had to dev a bunch of the other packages to get it to use 0.7 locally. It's a bit tricky because things like LayoutPointers need to have the 0.7 patch for ArrayInterface to get that working.

chriselrod:
> Sorry. I should have commented that I had to dev a bunch of the other packages to get it to use 0.7 locally. It's a bit tricky because things like LayoutPointers need to have the 0.7 patch for ArrayInterface to get that working.

Are you able to `using LoopVectorization`?
It will need updating, e.g. `pushexpr!` will result in method errors:
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/cbb9f2324ec3cb2c4aa5523d8c4fb98147aa7855/src/modeling/graphs.jl#L70
that should cause precompilation failure.

Tokazama:
You're right. That is causing issues. I'm not entirely sure what I did the first time but I can get it to compile now when I change that to `x::Union{Integer,StaticInt}`. I'm working on on the tests.

Tokazama:
Looks like I need to fix issues in StrideArraysCore before getting these to pass.

PR #426, opened 2022-08-31
Add a test for invalidation

timholy:
xref https://github.com/JuliaSIMD/CPUSummary.jl/issues/3#issuecomment-1232967989


chriselrod:
@timholy maybe there's a better way to run tests in parallel, but mind adding
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/def5ad12f760f0c97817082141a77c642159f4f4/.github/workflows/ci.yml#L36-L41
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/def5ad12f760f0c97817082141a77c642159f4f4/.github/workflows/ci-julia-nightly.yml#L36-L40

chriselrod:
```julia
  Activating project at `~/work/LoopVectorization.jl/LoopVectorization.jl/test/precompile`
ERROR: LoadError: ArgumentError: Package LoopVectorization [bdcacae8-1622-11e9-2a5c-532679323890] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.
```

timholy:
Thanks for adding it to CI, sorry I didn't get to it first.

DilumAluthge:
I would just add SnoopCompile and SnoopCompileCore as test dependencies, and get rid of the special "precompile" project.

DilumAluthge:
Why do you need to activate a separate project?

timholy:
LoopVectorization is used to build LVUser, so LVUser's environment has to be able to find LoopVectorization. And then we load the LVUser package and test whether this caused invalidations.

DilumAluthge:
Hmmm. Might be cleaner to do this in a subprocess. Let me give this a bit of thought.

timholy:
> what's the best way to make sure it finds the LV we're testing

This version may fix it :crossed_fingers: because in the Manifest for `LVUser` I specify the `path` for LoopVectorizations as `../../..`.

test/precompile/LVUser/src/LVUser.jl
```diff
@@ -0,0 +1,35 @@
...
+export filter2davx
+
+using LoopVectorization
+using SnoopPrecompile
+
+function filter2davx!(out::AbstractMatrix, A::AbstractMatrix, kern::AbstractMatrix)
+    @turbo for J in CartesianIndices(out)
+        tmp = zero(eltype(out))
+        for I ∈ CartesianIndices(kern)
+            tmp += A[I + J] * kern[I]
```
chriselrod, line 12:
```suggestion
            tmp += A[I + J - 1] * kern[I]
```
The maximum values of `I` and `J` are `size(kern)` and `size(out)`, so `I+J`'s maximum is `size(kern) .+ size(out) == size(kern) .+ size(A) .- size(kern) .+ 1 == size(A) .+ 1`, hence we get a bounds error on indexing `A`.
The minimum possible value is `2`.

Alternatively, you could use `OffsetArray`s.

This will hopefully fix those segfaults.

chriselrod:
Great, looks good, thank you!

Issue #430, opened 2022-09-14
Preventing `StackOverflowError` automatically with a `@safe_turbo`?


MilesCranmer:
Hey all,

I am eager to switch from `@inbounds @simd`  to `@turbo` in the evaluation loops of [SymbolicRegression.jl](https://github.com/MilesCranmer/SymbolicRegression.jl), which is the backend for [PySR](https://github.com/MilesCranmer/PySR). You can see my initial pull request here: https://github.com/MilesCranmer/SymbolicRegression.jl/pull/132 which shows the loops I am attempting to speed up.

The way my package works is that the user can pass any binary or unary operator (e.g., `+`, `-`, `*`, `/`, `cos`, `exp`, or any function they define). These operators will be arranged into expressions by a genetic algorithm until a combination is found that matches a relationship in a dataset. The evaluation of each expression needs to be high-performance, and is performed by a loop over an array with the `@inbounds @simd` macros. This loop is behind a function barrier so that the performance is the same as if I had hard-coded each operator the user passed.

I tried to switch to `@turbo`; however, I ended up seeing `StackOverflowError` when testing `SpecialFunctions.gamma` as an operator. From my look through these issues: https://github.com/JuliaSIMD/LoopVectorization.jl/issues/233, https://github.com/JuliaSIMD/LoopVectorization.jl/issues/232, it seems like certain functions will raise this error if they do not have SIMD implemented.

Now, this is an issue for my package, since the user is allowed (and encouraged) to specify any Julia function they wish as the operator - even complex, branching functions (including any of `SpecialFunctions`) - and those operators will be used inside these loops. Now, I understand that each operator would need to be implemented as a SIMD operation for `@turbo` to give a speedup, but I would still like to get the speed for other more standard operators, like `+`, `-`, `*`, `/`, etc. 

I am therefore wondering if it is possible to change `@turbo` to be robust against `StackOverflowErrors`, and fall back to a non-SIMD operations? Or, maybe a `@safe_turbo` macro implemented that could do this? (This assumes I do not know what code the user would use inside each loop at runtime.)

Thanks!
Miles

timholy:
A likely culprit is the assumption that constructors/conversion returns an object of the given type, see https://github.com/JuliaLang/julia/issues/42372

chriselrod:
Adding a `safe` key word arg to `@turbo` or a `@safe_turbo` option wouldn't be difficult.

If you want to try:
1. Add it to the kwargs here:
[...]

Issue #232, opened 2021-04-02
error with erf 


CarloLucibello:
Not sure this is the right place to ask (maybe VectorizationBase?) but this is what I get when I try to use `erf`

```julia
julia> using LoopVectorization, SpecialFunctions

julia> @avx for i=1:10
          a[i] = erf(x[i])
       end
ERROR: StackOverflowError:
Stacktrace:
 [1] erf(x::VectorizationBase.Vec{4, Float64}) (repeats 79984 times)
   @ SpecialFunctions ~/.julia/packages/SpecialFunctions/mFAQ4/src/erf.jl:10
```

Is `erf` support planned?

chriselrod:
> Is `erf` support planned?

Yes. Not sure what the best way to do this is, because LoopVectorization & VectorizationBase don't depend on SpecialFunctions. Maybe use `Requires.jl`, which LoopVectorization is already using anyway.

The [current implementation](https://github.com/JuliaSIMD/VectorizationBase.jl/blob/6404e0dae5bc5c27859347e5ebdfda991cae414f/src/llvm_intrin/verf.jl) needs work, but:
```julia
julia> using LoopVectorization, SpecialFunctions, VectorizationBase

julia> x = randn(10); a = similar(x);

julia> @avx for i=1:10
           a[i] = VectorizationBase.verf(x[i])
       end

julia> a'
1×10 adjoint(::Vector{Float64}) with eltype Float64:
 0.565029  -0.59615  -0.382111  0.429077  -0.879287  -0.272737  0.43213  -0.937869  -0.917235  0.224102

julia> erf.(x)'
1×10 adjoint(::Vector{Float64}) with eltype Float64:
 0.565029  -0.59615  -0.382111  0.429077  -0.879287  -0.272737  0.43213  -0.937869  -0.917235  0.224102
```
The current implementation consists of copy/pastes of LLVM emitted by compiling the xsimd C++ library using Clang.
But it's not cross platform, because that IR uses target specific intrinsics, so while it should work on anything x64, it will not work on ARM.
While that could be solved by compiling ARM versions...
The better plan would be to just write Julia versions. Then they could also benefit from accepting `VecUnroll` arguments, which can give a fairly good % performance boost to special functions thanks to out of order execution.

PR #431, opened 2022-09-18
Check operations in `@turbo` automatically with `can_avx`; if failure, switch to `@inbounds @fastmath`

MilesCranmer:
This solves #430 and prevents the StackOverflowError from appearing in #232.  cc @timholy @chriselrod. Thanks to @chriselrod for walking me through how to implement this with detailed instructions.

Basically this will check all instructions in a `@turbo` loop (if the `safe=true` kwarg is set - ~~`true`~~`false` by default) with `ArrayInterface.can_avx`. If any operation is `false`, then the check will fail and `@inbounds @fastmath` will be used instead, which will also print a warning message unless `warn_check_arg=false`.

```julia
using LoopVectorization
using SpecialFunctions

x = Float32.(1:0.1:10)
y = similar(x)

@turbo safe=true for i in indices(x)
    y[i] = gamma(x[i])
end
```
Note that ~~`safe=true`~~ `safe=false` by default ~~so this is technically not needed~~.

Before this change, this code would result in a mysterious `StackOverflowError`, since `gamma` does not have an AVX version implemented. 

As described in #430, this change is useful for cases where the user can pass an arbitrary operator - one still wants `@turbo` to work if that operator can be AVX'd, but default to `@inbounds @fastmath` otherwise.

---

Minor additional changes: I refactored `can_avx.jl`, and included `SpecialFunctions.gamma` as an operator which cannot currently be AVX'd.


MilesCranmer:
I'm not sure if `can_avx` is as general as one would like? There are some operators which can `@turbo` fine, but `can_avx` fails. e.g., if I just write:
```
f(x) = exp(x)
can_avx(f)
```
this would return `false`. Is there a more general way to check this @chriselrod? I suppose I should turn `safe=false` by default anyways, just in case situations like this occur.

chriselrod:
> Is there a more general way to check this @chriselrod?

https://github.com/JuliaSIMD/LoopVectorization.jl/issues/430#issuecomment-1247256476
You could use `promote_op` as described there, checking the number of arguments used. 
Getting the element types would be harder, but you could assume `Vec{2,Int}` if you don't want to do anything fancy.

chriselrod:
For example:
```julia
julia> using SpecialFunctions, VectorizationBase, SLEEFPirates

julia> can_turbo(f::F, ::Val{NARGS}) where {F,NARGS} = Base.promote_op(f, ntuple(Returns(Vec{2,Int}), Val(NARGS))...) !== Union{}
can_turbo (generic function with 1 method)

julia> can_turbo(+, Val(1))
true

julia> can_turbo(exp, Val(1))
true

julia> f(x) = exp(x)
f (generic function with 1 method)

julia> can_turbo(f, Val(1))
true

julia> can_turbo(gamma, Val(1))
false
```
This isn't necessarilly precise, but should work well in practice.

MilesCranmer:
What is the best way to get `NARGS`? Maybe `length(op.dependencies)`?

chriselrod:
> What is the best way to get `NARGS`? Maybe `length(op.dependencies)`?

`length(parents(op))`

MilesCranmer:
Awesome, thanks! Everything is implemented, and added a few tests too. Let me know what you think.

Also, let me know if you'd rather have `safe=true` by default. Right now it is `false` since I wasn't sure how general this technique is.

MilesCranmer:
I'm not sure why the tests are failing - did I miss a `, safe` snippet somewhere?

(probably would be good to refactor away the shotgun surgery in the future, if possible)

chriselrod:
```julia
broadcast_float: Error During Test at /home/runner/work/LoopVectorization.jl/LoopVectorization.jl/test/broadcast.jl:205
[243](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:246)
  Got exception outside of a @test
[244](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:247)
  BoundsError: attempt to access Tuple{Bool, Int8, Int8, Int8, Bool, Int64, Int64, Int64, Int64, UInt64} at index [11]
[245](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:248)
  Stacktrace:
[246](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:249)
    [1] indexed_iterate(t::Tuple{Bool, Int8, Int8, Int8, Bool, Int64, Int64, Int64, Int64, UInt64}, i::Int64, state::Int64)
[247](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:250)
      @ Base ./tuple.jl:88
[248](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:251)
    [2] #s217#153
[249](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:252)
      @ ~/work/LoopVectorization.jl/LoopVectorization.jl/src/broadcast.jl:551 [inlined]
[250](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:253)
    [3] var"#s217#153"(T::Any, N::Any, BC::Any, Mod::Any, UNROLL::Any, dontbc::Any, ::Any, dest::Any, bc::Any, #unused#::Type, #unused#::Type, #unused#::Any)
[251](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:254)
      @ LoopVectorization ./none:0
[252](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:255)
    [4] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any})
[253](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:256)
      @ Core ./boot.jl:597
[254](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:257)
    [5] vmaterialize
[255](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3078404283/jobs/4974198686#step:6:258)
      @ ~/work/LoopVectorization.jl/LoopVectorization.jl/src/broadcast.jl:679 [inlined]
```
Looks like you didn't update broadcasting.

chriselrod:
Forward it here:
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/88dfe72e29a7308d551f894b8cf8cc2ceaf21404/src/constructors.jl#L46-L60

chriselrod:
Okay, `avx_config_val` is a bit of a mess.
It's defined in `condense_loopset.jl`, has a different length depending on broadcast vs not.

Currently, it's dropping `safe`, hence it's missing in the broadcast files, resulting in the indexing errors.

MilesCranmer:
I tried to add `warncheckarg` and `safe` everywhere the packing/unpacking occurs... still not sure where the other errors are coming from. It's really hard to debug this message:
```
    [1] indexed_iterate(t::Tuple{Bool, Int8, Int8, Int8, Bool, Int64, Int64, Int64, Int64, UInt64}, i::Int64, state::Int64)
      @ Base ./tuple.jl:88
    [2] #s191#153
      @ ~/Documents/LoopVectorization.jl/src/broadcast.jl:551 [inlined]
    [3] var"#s191#153"(T::Any, N::Any, BC::Any, Mod::Any, UNROLL::Any, dontbc::Any, ::Any, dest::Any, bc::Any, #unused#::Type, #unused#::Type, #unused#::Any)
      @ LoopVectorization ./none:0
    [4] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any})
      @ Core ./boot.jl:582
    [5] vmaterialize!(dest::OffsetVector{Float64, Vector{Float64}}, bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(identity), Tuple{OffsetVector{Float64, Vector{Float64}}}}, #unused#::Val{:Main}, #unused#::Val{(true, 0, 0, 0, true, 0, 16, 31, 128, 0x0000000000000001)})
      @ LoopVectorization ~/Documents/LoopVectorization.jl/src/broadcast.jl:666
    [6] macro expansion
      @ ~/Documents/LoopVectorization.jl/test/iteration_bound_tests.jl:19 [inlined]
    [7] macro expansion
      @ /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/Test/src/Test.jl:1357 [inlined]
```

Traceback `[6]` is just a call with `@turbo` and then `[5]` is `vmaterialize` - I have no idea where the `UNROLL` actually gets packed...

MilesCranmer:
Okay, I think I found them all. (I would definitely recommend making a dedicated type for `UNROLL` instead of unpacking/packing (maybe a `NamedTuple`?) -  I think the current pattern will inevitably cause bugs.)

Let me know how this PR looks.
Thanks,
Miles

chriselrod:
> Okay, I think I found them all. (I would definitely recommend making a dedicated type for `UNROLL` instead of unpacking/packing (maybe a `NamedTuple`?) - I think the current pattern will inevitably cause bugs.)
> 
> Let me know how this PR looks. Thanks, Miles

Feel free to add this yourself =)

MilesCranmer:
> Feel free to make this change yourself =)

Not sure I have enough knowledge of LoopVectorization.jl to do such a big refactor - maybe one day when I have more time :)

test/safe_turbo.jl
```diff
@@ -16,6 +16,7 @@
   # `gamma` can be AVX'd, but `can_turbo` can:
   f(x) = exp(x)
 
+  @test LoopVectorization.ArrayInterface.can_avx(f)
```
chriselrod, line 19: was this supposed to be `can_avx(exp)`?

MilesCranmer: Yes, nice (fast!) catch!

chriselrod:
```julia
 shuffles load/stores     | 4442    257   4699  1m15.4s
```
These nightly failures happen on the `main` branch, too.

chriselrod:
Add special functions to `test/Project.toml`

MilesCranmer:
Seeing this issue now:
```julia
  Test threw exception
  Expression: LoopVectorization.can_turbo(f2, Val(1))
  UndefVarError: Returns not defined
  Stacktrace:
   [1] can_turbo(f::var"#f2#2", #unused#::Val{1})
     @ LoopVectorization ~/work/LoopVectorization.jl/LoopVectorization.jl/src/condense_loopset.jl:913
   [2] macro expansion
     @ ~/work/LoopVectorization.jl/LoopVectorization.jl/test/safe_turbo.jl:21 [inlined]
   [3] macro expansion
     @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
   [4] top-level scope
     @ ~/work/LoopVectorization.jl/LoopVectorization.jl/test/safe_turbo.jl:4
```
I guess this is from the `Returns(Vec{2,Int})` snippet? Not sure how to fix this

MilesCranmer:
I'm traveling for the rest of this week unfortunately, so I won't be online much. I can retry this weekend though. Let me know if you have any clues as to why this breaks. Also, feel free to push to the branch (and potentially merge) as you wish.

chriselrod:
> I guess this is from the `Returns(Vec{2,Int})` snippet? Not sure how to fix this

`Returns` was new in Julia 1.7; it didn't exist in Julia 1.6.

src/condense_loopset.jl
```diff
@@ -885,6 +903,38 @@ function check_args_call(ls::LoopSet)
   q
 end
```
chriselrod, line 905:
```suggestion
struct RetVec2Int end
(::RetVec2Int)(_) = Vec{2,Int}
```

src/condense_loopset.jl
```diff
@@ -885,6 +903,38 @@ function check_args_call(ls::LoopSet)
...
 
+"""
+  can_turbo(f::Function, ::Val{NARGS})
+
+Check whether a given function with a specified number of arguments
+can be used inside a `@turbo` loop.
+"""
+function can_turbo(f::F, ::Val{NARGS})::Bool where {F,NARGS}
+  promoted_op = Base.promote_op(f, ntuple(Returns(Vec{2,Int}), Val(NARGS))...)
```
chriselrod, line 913:
```suggestion
  promoted_op = Base.promote_op(f, ntuple(RetVec2Int(), Val(NARGS))...)
```

chriselrod (commented):
This is an example workaround for `Returns` not existing yet.

MilesCranmer:
I couldn't figure that one out, was hoping you would have some idea. I tried defining the function inside and outside of the test set, and ensuring that each function has a unique name, but that error persists. I assume it's some weird test syntax quirk.

MilesCranmer:
Nice!! Thanks

chriselrod:
Thanks for all the work on this!

PR #439, opened 2022-10-12
Length of tuple

chriselrod:
this code wasn't reached anyway, so perhaps it should be removed?


src/condense_loopset.jl
```diff
@@ -919,6 +919,7 @@ can_turbo(::typeof(vfmsubadd), ::Val{3}) = true
 can_turbo(::typeof(ifelse), ::Val{3}) = true
 can_turbo(::typeof(!), ::Val{1}) = true
 can_turbo(::typeof(^), ::Val{2}) = true
+can_turbo(::typeof(Base.FastMath.pow_fast), ::Val{2}) = true
```
chriselrod, line 922:
```suggestion
```

PR #444, opened 2022-10-31
Add DynamicExpressions.jl and PySR/SymbolicRegression.jl to example packages

MilesCranmer:
With new versions of these packages, there's a `turbo` mode for faster expression evaluation!


chriselrod:
Thanks for your contributions to get `@turbo` to be safer!

PR #445, opened 2022-11-21
let homogeneous tuples be `alldense`

JeffreySarnoff:
This lets LoopVectorization work more broadly .. e.g. VectorizedStatistics can process NTuples, `vminimum((1.0,2.0,0.5,1.5))` (currently it chokes).

### DIFF
--- a/src/simdfunctionals/map.jl
+++ b/src/simdfunctionals/map.jl
@@ -266,7 +266,7 @@ end
 @inline _all_dense(t::Tuple{ArrayInterface.True,ArrayInterface.False,Vararg}) = false
 @inline _all_dense(t::Tuple{ArrayInterface.False,Vararg}) = false
 @inline all_dense() = true
-@inline all_dense(t::NTuple{N,T}) where {N,T} = true
+@inline all_dense(t::NTuple{N}) where {N} = true
 @inline all_dense(A::Array) = true
 @inline all_dense(A::AbstractArray) = _all_dense(ArrayInterface.dense_dims(A))
 @inline all_dense(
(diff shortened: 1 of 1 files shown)


JeffreySarnoff:
clearly I overlooked something structural -- closing this PR 
will open an issue covering the same thing

src/simdfunctionals/map.jl
```diff
@@ -266,6 +266,7 @@ end
 @inline _all_dense(t::Tuple{ArrayInterface.True,ArrayInterface.False,Vararg}) = false
 @inline _all_dense(t::Tuple{ArrayInterface.False,Vararg}) = false
 @inline all_dense() = true
+@inline all_dense(t::NTuple{N,T}) where {N,T} = true
```
chriselrod, line 269:
```suggestion
@inline all_dense(t::NTuple{N}) where {N} = true
```
should fix the test failures. An `NTuple` is already homogenous. If `N == 0`, then `T` would be an unbound arg.

chriselrod:
> clearly I overlooked something structural -- closing this PR will open an issue covering the same thing

What did you overlook? Is this not enough for the example `vminimum` to work?

JeffreySarnoff:
It does make the v_ functions work -- what I overlooked is whatever is causing the test to fail here
https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3512452079/jobs/5884165315
If that is unrelated -- cool.

chriselrod:
> It does make the v_ functions work -- what I overlooked is whatever is causing the test to fail here https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/3512452079/jobs/5884165315 If that is unrelated -- cool.

That is fixed by my suggestion above.

chriselrod:
I went ahead and committed my suggestion; I'll merge if tests pass.

PR #448, opened 2022-12-03
fix some invalidations caused by FixedPointNumbers

ranocha:
This is based on the following code (on Julia v1.8.3):
```julia
julia> using Pkg; Pkg.activate(temp=true); Pkg.add("FixedPointNumbers"); Pkg.develop("LoopVectorization")

julia> using SnoopCompileCore; invalidations = @snoopr(using LoopVectorization, FixedPointNumbers); using SnoopCompile

julia> length(uinvalidated(invalidations))
613

julia> trees = invalidation_trees(invalidations);

julia> trees[end]
inserting reduce_first(::typeof(Base.add_sum), x::FixedPoint) in FixedPointNumbers at ~/.julia/packages/FixedPointNumbers/HAGk2/src/FixedPointNumbers.jl:295 invalidated:
   backedges: 1: superseding reduce_first(::typeof(Base.add_sum), x) in Base at reduce.jl:402 with MethodInstance for Base.reduce_first(::typeof(Base.add_sum), ::Any) (166 children)

julia> ascend(trees[end].backedges[end])
```

With this PR, 
```julia
julia> length(uinvalidated(invalidations))
513
```

The initial number of invalidations coming from this is bigger when loading something doing more precompilation involving LoopVectorization.jl.

### DIFF
--- a/src/modeling/graphs.jl
+++ b/src/modeling/graphs.jl
@@ -834,7 +834,8 @@ function add_constant_compute!(ls::LoopSet, op::Operation, var::Symbol)::Operati
     if !getconstfailed
       f = instr.instr
       if f === :add_fast
-        return add_constant!(ls, sum(vals), 8)::Operation
+        # use init explicitly to prevent invalidations
+        return add_constant!(ls, sum(vals; init = false), 8)::Operation
       elseif f === :mul_fast
         return add_constant!(ls, prod(vals), 8)::Operation
       elseif f === :sub_fast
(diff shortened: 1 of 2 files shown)


chriselrod:
Would making `vals` a `Number[]` instead of an `Any[]` also help at all?

chriselrod:
The test failures are likely related to `StrideArraysCore v0.4.0`. I pushed it out prematurely.

Neither the DiffEq stack or StrideArrays.jl have releases supporting it yet (for good reason), so it's mostly been unnoticed.

ranocha:
> Would making `vals` a `Number[]` instead of an `Any[]` also help at all?

No, I don't think so. The invalidations come from FixedPointNumbers.jl, which provides new `Number` types with specializations.

ranocha:
@chriselrod Would you mind making a new release of LoopVectorization.jl including these invalidation fixes?

PR #451, opened 2022-12-28
Use 16 bits to store an array reference index

johnbcoughlin:
This seems to work without causing any extra errors, hopefully CI bears it out.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LoopVectorization"
 uuid = "bdcacae8-1622-11e9-2a5c-532679323890"
 authors = ["Chris Elrod <elrodc@gmail.com>"]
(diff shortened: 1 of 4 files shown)


chriselrod:
Thanks. I guess you've confirmed this fixes your problem locally?

If it's not too much trouble, mind adding a test?

Also, please feel free to bump the patch version so I can issue a new release after merging.

johnbcoughlin:
Done. The new test fails on master with the same error. Let me know if you'd like me to test a more general property now that the 8 bit limit is raised.

chriselrod (approved):
Thanks, looks great!

chriselrod:
Failures are unrelated to this PR.

PR #456, opened 2023-01-05
Fix invalidations from new `==` method

timholy:
```
julia> tree
inserting ==(retcode::SciMLBase.ReturnCode.T, s::Symbol) @ SciMLBase ~/.julia/packages/SciMLBase/QqtZA/src/retcodes.jl:348 invalidated:
   mt_backedges: 1: signature ==(x, y) @ Base Base.jl:127 (formerly ==(x, y) @ Base Base.jl:127) triggered MethodInstance for MacroTools.store!(::Dict{Any, Any}, ::Symbol, ::Symbol) (3 children)
                 2: signature ==(x, y) @ Base Base.jl:127 (formerly ==(x, y) @ Base Base.jl:127) triggered MethodInstance for isequal(::Any, ::Symbol) (1 children)
   backedges: 1: superseding ==(x, y) @ Base Base.jl:127 with MethodInstance for ==(::Any, ::Symbol) (224 children)
   2 mt_cache
```

The plurality of those invalidations are LoopVectorization-related.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -201,13 +201,13 @@ function replace_enumerate!(q, prepreamble)
 end
 function replace_single_enumerate!(q, prepreamble, i = nothing)
   if isnothing(i) # not nest loop
-    looprange, body = q.args[1], q.args[2]
+    looprange, body = q.args[1]::Expr, q.args[2]
   else # nest loop
-    looprange, body = q.args[1].args[i], q.args[2]
+    looprange, body = q.args[1].args[i]::Expr, q.args[2]
   end
   @assert Meta.isexpr(looprange, :(=), 2)
   itersyms, r = looprange.args
-  if Meta.isexpr(r, :call, 2) && r.args[1] == :enumerate
+  if Meta.isexpr(r, :call, 2) && r.args[1] === :enumerate
     _iter = r.args[2]
     if _iter isa Symbol
       iter = _iter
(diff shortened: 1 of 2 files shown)


chriselrod (approved):
Thanks for the fixes as always!
I'm surprised there was still an `==` left.

```
> git grep -nr " == :" src
src/constructors.jl:210:  if Meta.isexpr(r, :call, 2) && r.args[1] == :enumerate
src/vectorizationbase_compat/contract_pass.jl:88:    (fun == :(Base.FastMath.add_fast))::Bool
src/vectorizationbase_compat/contract_pass.jl:93:    (fun == :(Base.FastMath.sub_fast))::Bool
src/vectorizationbase_compat/contract_pass.jl:117:         (f == :(Base.FastMath.mul_fast))::Bool
```

chriselrod:
Feel free to bump the version.

chriselrod:
Would it be too crazy to call internal functions that have no business being invalidated through `invoke_in_world`?

chriselrod:
> > Would it be too crazy to call internal functions that have no business being invalidated through invoke_in_world?
> 
> Maybe not? It might be a bit uglier. I kind think we want to be able to annotate methods and callsites with `@concrete_dispatch`: the rule being that if all the types infer concretely, just use normal dispatch, but if any of them are abstract then use runtime dispatch. This would prevent us from using abstract inference, which is the main invalidation risk. Come to think of it, maybe one wants to be able to make that a module-wide setting?

What's the difference between that and `max_methods=1`?

PR #462, opened 2023-01-24
Unconditional max_methods=1

chriselrod:
@ChrisRackauckas no fear
`max_methods=1` can't hurt you if ~~you're pure of heart~~ type stable.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "LoopVectorization"
 uuid = "bdcacae8-1622-11e9-2a5c-532679323890"
 authors = ["Chris Elrod <elrodc@gmail.com>"]
-version = "0.12.147"
+version = "0.12.148"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
--- a/src/LoopVectorization.jl
+++ b/src/LoopVectorization.jl
@@ -1,5 +1,10 @@
 module LoopVectorization
 
+if isdefined(Base, :Experimental) &&
+   isdefined(Base.Experimental, Symbol("@max_methods"))
+  @eval Base.Experimental.@max_methods 1
+end
+
 using ArrayInterfaceCore: UpTri, LoTri
 using Static: StaticInt, gt, static, Zero, One, reduce_tup
 using VectorizationBase,


chriselrod:
Well, it doesn't seem to help either.
PR:
```julia
julia> @time @eval function AmulB!(C,A,B)
           @turbo for n ∈ indices((C,B),2), m ∈ indices((C,A),1)
               Cmn = zero(eltype(C))
               for k ∈ indices((A,B),(2,1))
                   Cmn += A[m,k] * B[k,n]
               end
               C[m,n] = Cmn
           end
       end
  0.023787 seconds (4.70 k allocations: 271.455 KiB, 76.64% compilation time)
AmulB! (generic function with 1 method)

julia> M = K = N = 71; A = rand(M,K); B = rand(K,N); C1 = A*B; C0 = zero(C1);

julia> @time @eval AmulB!(C0, A, B);
  2.113024 seconds (4.32 M allocations: 261.378 MiB, 3.93% gc time, 99.99% compilation time)
```
Latest release:
```julia
julia> @time @eval function AmulB!(C,A,B)
           @turbo for n ∈ indices((C,B),2), m ∈ indices((C,A),1)
               Cmn = zero(eltype(C))
               for k ∈ indices((A,B),(2,1))
                   Cmn += A[m,k] * B[k,n]
               end
               C[m,n] = Cmn
           end
       end
  0.024604 seconds (4.70 k allocations: 271.307 KiB, 76.93% compilation time)
AmulB! (generic function with 1 method)

julia> M = K = N = 71; A = rand(M,K); B = rand(K,N); C1 = A*B; C0 = zero(C1);

julia> @time @eval AmulB!(C0, A, B);
  2.093600 seconds (4.30 M allocations: 260.192 MiB, 4.07% gc time, 99.99% compilation time)
```
Of course, there is a ton of type unstable code in LV to process expressions.

Maybe I should merge this anyway as a symbolic gesture.

ChrisRackauckas:
I was wondering if it could effect compilation downstream though when it's being hit behind a function barrier (due to the autoswitch solvers). That's really tedious to test though.

chriselrod:
> Of course, there is a ton of type unstable code in LV to process expressions.

Which is why I was thinking this may help.

> I was wondering if it could effect compilation downstream though when it's being hit behind a function barrier (due to the autoswitch solvers). That's really tedious to test though.

I'm going to `@descend` through the calls in OrdinaryDiffEq's precompile statements and see if I run into instabilities.

chriselrod:
> I'm going to `@descend` through the calls in OrdinaryDiffEq's precompile statements and see if I run into instabilities.

I run into type instabilities immediately.

ChrisRackauckas:
There should be some due to the auto-switch, which does manual branching in order to get around the way it would dispatch, but I'm not sure it gives good precompilation.

chriselrod:
Even without autoswitch

PR #469, opened 2023-02-20
Updates for StaticArrayInterface

chriselrod:


chriselrod:
Those piracy failures are real, and my bad. I own the libraries being pirated though. I places the extensions here for historical reasons, but they really should have gone to the owning libraries.

For now, that is fine. May change later.

PR #494, opened 2023-05-18
@turbo function(;kwag) work around

alex-s-gardner:
I've added a note that @turbo can not be applied to functions that are passed kwargs and provided an example of how to work around this limitation. Notes are provided in both the `ReadMe.md` and the `getting_started.md`


chriselrod:
Thanks

PR #496, opened 2023-05-21
Create `vsum`

MilesCranmer:
This creates a vectorized version of `sum`, as has been done with `vmapreduce` and `vreduce` already. The implementation is just an alias:

```julia
@inline vsum(f::F, A::AbstractArray{T}) where {F,T<:NativeTypes} = vmapreduce(f, +, A)
@inline vsum(A::AbstractArray{T}) where {T<:NativeTypes} = vsum(identity, A)
```

(There is a `vsum` in VectorizationBase.jl; that one is I think (?) an internal utility, as it doesn't take normal arrays.)

For example:

```julia

julia> x = rand(127);

julia> @btime vsum(hypot, $x)
  12.095 ns (0 allocations: 0 bytes)
66.65246070098374

julia> @btime sum(hypot, $x)
  16.992 ns (0 allocations: 0 bytes)
66.65246070098372
```

### DIFF
--- a/docs/src/vectorized_convenience_functions.md
+++ b/docs/src/vectorized_convenience_functions.md
@@ -132,4 +132,22 @@ julia> @btime mapreduce(hypot, +, $x, $y)
 96.75538300513509
 ```
 
+## vsum
+
+Vectorized version of `sum`. `vsum(f, a)` applies `f(a[i])` for `i in eachindex(a)`, then sums the results.
+
+```julia
+julia> using LoopVectorization, BenchmarkTools
+
+julia> x = rand(127);
+
+julia> @btime vsum(hypot, $x)
+  12.095 ns (0 allocations: 0 bytes)
+66.65246070098374
+
+julia> @btime sum(hypot, $x)
+  16.992 ns (0 allocations: 0 bytes)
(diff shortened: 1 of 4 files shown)


chriselrod (approved):
Sure.
If people want much more than this, it might also be good to point people to:
https://github.com/andrewjradcliffe/VectorizedReduction.jl
https://github.com/JuliaSIMD/VectorizedStatistics.jl

chriselrod:
You need to export `vsum` if you're testing that it is exported.
The `oftype` failures on nightly are unrelated to this PR.

MilesCranmer:
Fixed, thanks.

That package looks super useful, I didn’t know about it. It might be a good idea to ask and try to bring it under the JuliaSIMD organization, for consistency and long term support. I think having it seem like a more “official” package would make me more likely to use it.

MilesCranmer:
Oh I didn’t know about VectorizedStatistics! Should I close this PR?

chriselrod:
`vsum` is simple enough so that it's fine. But maybe we shouldn't export it if they do.

@andrewjradcliffe you're welcome to move your package to JuliaSIMD if you'd like.

PR #503, opened 2023-06-29
Fix typos and update hyperlinks

alexfreudenberg:
Maybe relative links to internal files would help to avoid breaking them?


chriselrod:
Thanks!
But in this case the problem was that the files changed directories, e.g. moving from `src` to `src/modeling`. This break would not have been avoided by making the paths relative.

PR #516, opened 2023-12-06
Separate `LoopVectorization.vsum` from `VectorizationBase.vsum`

brenhinkeller:
Seems like it might be better to have `LoopVectorization.vsum != VectorizationBase.vsum` esp. if we might want to extend the former (but not the latter) in other packages. I think this isn't technically breaking and hopefully it doesn't break anything if I did it right, but will see how tests look

### DIFF
--- a/src/LoopVectorization.jl
+++ b/src/LoopVectorization.jl
@@ -82,7 +82,6 @@ using VectorizationBase:
   reduced_any,
   reduce_to_all,
   reduce_to_any,
-  vsum,
   vprod,
   vmaximum,
   vminimum,
--- a/src/modeling/costs.jl
+++ b/src/modeling/costs.jl
@@ -598,7 +598,7 @@ reduce_number_of_vectors(x::Float64) =
   end
 reduction_to_scalar(x::Float64) =
   if x == ADDITIVE_IN_REDUCTIONS
-    :vsum
+    :(VectorizationBase.vsum)
   elseif x == MULTIPLICATIVE_IN_REDUCTIONS
     :vprod
   elseif x == MAX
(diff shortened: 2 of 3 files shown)


chriselrod:
Thanks for addressing this/taking over from #515.
Note that LoopVectorization itself calls `VectorizationBase.vsum`.
Double check whether it namespacing them properly (`VectorizationBase.vsum`), or being lazy (`LoopVectorization.vsum` and relying on the `using VectorizationBase: vsum` that you removed here).

brenhinkeller:
Oh oops I hadn’t noticed #515! Looks like this is still breaking something, but I don't think I understand why..

chriselrod:
> Oh oops I hadn’t noticed #515! Looks like this is still breaking something, but I don't think I understand why..

Some failures are because LV uses VectorizationBase.vsum, without saying VectorizationBase.vsum.

chriselrod:
> That would make sense but I can't find any other usages of `vsum` when I search the package...


E.g.
https://github.com/JuliaSIMD/LoopVectorization.jl/blob/d2f749dbf95c161c803f8733e0b4376c441c3a4a/src/modeling/costs.jl#L599-L601

https://github.com/JuliaSIMD/LoopVectorization.jl/blob/d2f749dbf95c161c803f8733e0b4376c441c3a4a/src/modeling/operations.jl#L564-L565

LV doesn't call the code, it generates expressions with calls.

brenhinkeller:
Ooh, so is it as simple as replacing that with `:(VectorizationBase.vsum)` or such? Trying that locally now and looks a bit better but still some local test failures

chriselrod:
> Ooh, so is it as simple as replacing that with `:(VectorizationBase.vsum)` or such? Trying that locally now and looks a bit better but still some local test failures

Note that `lv(x)` (called in one of those code snippets) does `GlobalRef(LoopVectorization, x)`. We need VectorizationBase instead.

PR #519, opened 2024-01-03
Deprecate LV for Julia >= 1.11-DEV

chriselrod:


chriselrod:
Fixes #518

stillyslalom:
It's sad to see this project get mothballed, but I understand not wanting to deal with the maintenance burden when you're trying to get LoopModels up to parity. Do you think LoopModels will be usable in the v1.11-release timeframe, or is it still too early in its lifecycle to guess on a release date?

willow-ahrens:
I agree, this package will be sorely missed. Thank you Chris for your work on this package, it is quite an achievement.

For future implementers, would you share any ideas towards a more minimal version of the package with less maintenance burden and less compile time? Something that e.g. faithfully unrolls the inner loop, manually applies simd instructions, and tries different loop permutations without getting too serious about it?

chriselrod:
> Something that e.g. faithfully unrolls the inner loop

LV will often unroll and SIMD one of the outer loops, not the inner most!

chriselrod:
> It's sad to see this project get mothballed, but I understand not wanting to deal with the maintenance burden when you're trying to get LoopModels up to parity. Do you think LoopModels will be usable in the v1.11-release timeframe, or is it still too early in its lifecycle to guess on a release date?

It's too early to guess on a release date, but I would not except it by Julia 1.11.

PR #521, opened 2024-01-04
remove debug statements

ranocha:
These debug statements were added in https://github.com/JuliaSIMD/LoopVectorization.jl/pull/519. They made CI fail in https://github.com/ranocha/SummationByPartsOperators.jl/pull/241. I think it would be nice to disable them again (or only enable them with a special debug setting?).
It would be nice to get a new release when this is merged.


chriselrod:
Oops, I did not mean to add those.

ranocha:
Thanks for merging this so quickly!

PR #522, opened 2024-01-05
bump patch version

ranocha:
It would be nice to get a new release after merging #521


chriselrod:
Given https://github.com/JuliaRegistries/General/pull/98171 you shouldn't be seeing those print statements I accidentally committed when using the latest release, which makes this less urgent.

So following https://github.com/JuliaSIMD/LoopVectorization.jl/issues/520 might just release this as 0.13, but if packages need to update, they may as well drop `@turbo` and LV instead in preparation for the future.

PR #557, opened 2025-09-01
Updating LoopVectorization for 1.12

mxpoch:
Adding changes to LV for compatibility with 1.12

Done as part of my contributions for the [SciML small grants initiative](https://github.com/SciML/sciml.ai/blob/master/small_grants.md#update-loopvectorization-to-support-changes-in-julia-v112-200)


Project.toml
```diff
@@ -29,10 +29,11 @@ VectorizationBase = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
 [weakdeps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 
 [extensions]
-ForwardDiffExt = ["ChainRulesCore", "ForwardDiff"]
+ForwardDiffExt = ["ChainRulesCore", "ForwardDiff", "NNlib"]
```
ChrisRackauckas, line 36: NNLib should be a separate extension. Not only is this breaking, it also means that it won't properly trigger for most users.

ext/ForwardDiffExt.jl
```diff
@@ -157,6 +158,27 @@ end
   end
 end
 
+@generated function NNlib.leakyrelu(
+  x::ForwardDiff.Dual{T,S,N},
```
ChrisRackauckas, line 162: This is type piracy. Why is this required?

chriselrod: Those functions were once written with branches.
LV only added methods for the SIMD types and Duals of SIMD types where this was problematic. 
Not sure why this PR dropped the constraints.

mxpoch: Because that's what the [test suite tested for](https://github.com/JuliaSIMD/LoopVectorization.jl/blob/945c2f7b92df094cafe91ec29d56ee2fc29c5726/test/forwarddiffext.jl#L49) and was causing [this error](https://github.com/JuliaSIMD/LoopVectorization.jl/actions/runs/17360780179/job/49280598809?pr=556#step:6:5943) -- which occurs across pre, 1, and lts. I figured it would be cleaner to move the definition to the extension instead of redefining it in the [test](https://github.com/JuliaSIMD/LoopVectorization.jl/blob/945c2f7b92df094cafe91ec29d56ee2fc29c5726/test/forwarddiffext.jl#L26).

ChrisRackauckas:
Because `relu` is not from LoopVec anymore, it needs to be careful to not be pirating, while not disabling the adding of other dispatches. The way you had it, not adding NNlib would disable all ForwardDiff overloads, and it would pirate `NNlib.relu`'s `rrule` when it is added. I made a few code movements so that those bad behaviors are avoided.

It is still a bit odd that the rrules for SLEEFPirates are in there since they have nothing to do with ForwardDiff, but they at least were there before so this is at least not a regression from the previous behavior.

ChrisRackauckas:
To me, if this passes tests then it's good to go.

ChrisRackauckas:
Open a PR to the small grants to claim your finish.

chriselrod:
Congrats on the PR!

As an aside that is unrelated to this PR, I looked at an assembly example from LV in the past week and realized the quality regressed since a year ago or so.
This is probably just the inevitable decay as LLVM changes, and some of the tricks for getting good code gen out of it fall out of date.

Issue #543, opened 2025-02-14
No method matching `_vstore_unroll!` on ARM


benegee:
We are using `@turbo` extensively in [Trixi.jl](https://github.com/trixi-framework/Trixi.jl). 
Recently, we have started running our code on ARM-based machines and encountered the following error:

```
LoadError: MethodError: no method matching _vstore_unroll!(::LayoutPointers.StridedPointer{Float64, 4, 1, 0, (1, 2, 3, 4), Tuple{Static.StaticInt{8}, Static.StaticInt{8}, Static.StaticInt{40}, Static.StaticInt{200}}, NTuple{4, Static.StaticInt{0}}}, ::VectorizationBase.VecUnroll{4, 1, Float64, VectorizationBase.VecUnroll{4, 1, Float64, Float64}}, ::VectorizationBase.Unroll{2, 1, 5, 1, 1, 0x0000000000000000, 1, VectorizationBase.Unroll{4, 1, 5, 1, 1, 0x0000000000000000, 1, Static.StaticInt{0}}}, ::Static.False, ::Static.False, ::Static.False, ::Static.StaticInt{16}, ::Static.StaticInt{8})
  
  Closest candidates are:
    _vstore_unroll!(::LayoutPointers.AbstractStridedPointer{T1, D, C, B, R, X, O} where {B, R, X<:Tuple{Vararg{Union{Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8, Static.StaticInt}, D}}, O<:Tuple{Vararg{Union{Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8, Static.StaticInt}, D}}}, ::VectorizationBase.VecUnroll{<:Any, W, T2, <:VectorizationBase.VecUnroll{<:Any, W, T2, VectorizationBase.Vec{W, T2}}}, ::UU, ::M, ::A, ::S, ::NT, ::Static.StaticInt{RS}, ::SVUS) where {T1, D, C, W, T2, UU, A, S, NT, RS, SVUS, M}
     @ VectorizationBase ~/.julia/packages/VectorizationBase/LqJbS/src/vecunroll/memory.jl:2552
[...]

PR #562, opened 2025-11-28
add new test for issue 543

ranocha:
I added a fix for #543 in https://github.com/JuliaSIMD/VectorizationBase.jl/pull/125. However, Chris Elrod mentioned that this fix is not really good; instead, the fix should avoid the issue appearing there altogether, likely by fixing an issue inside LoopVectorization.jl. Thus, I added tests for the issue here that pass on Windows/Linux CI and are skipped on Apple ARM systems since they throw a `MethodError`.

### DIFF
--- a/test/staticsize.jl
+++ b/test/staticsize.jl
@@ -161,28 +161,32 @@ function issue543_turbo!(data_out, matrix, data_in)
 end
 
 @testset "Issue #543: W=1 Nested VecUnroll" begin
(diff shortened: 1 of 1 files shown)


ranocha:
@ChrisRackauckas This is similar to https://github.com/JuliaSIMD/LoopVectorization.jl/pull/563 since it discovers some tests that fail on MacOS ARM systems. Can we merge it, too, to keep tracking it in https://github.com/JuliaSIMD/LoopVectorization.jl/issues/564?

